PR #17, opened 2016-08-05
Lazy evaluation and no segmentation fault when reading str arrays

laliberte:
A few fixes.
1) both groups and variables can be lazily evaluated and,
2) no h5py object is stored in subsequent classes. This feature was causing segmentation faults when reading arrays with str datatype.

The changes are fairly minor with only a few additions to the hidden API.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -62,11 +62,10 @@ exceptions of:
 
 We simply haven't gotten around to implementing these features yet.
 
-A new feature, enabled by default, is the lazy opening of datasets. Unlike
+A new feature is the lazy opening of datasets. Unlike
 the netCDF interface where attributes for all variables and groups are read at
 the creation of the Dataset object, h5netcdf loads these attributes only
-when accessed. Passing `--lazy=False` to `h5netcdf.File` or `h5netcdf.legacyapi.Dataset`
-reverts to the netCDF interface behavior.
+when accessed.
 
 New API
 ~~~~~~~
--- a/README.rst
+++ b/README.rst
@@ -62,10 +62,14 @@ exceptions of:
 
 We simply haven't gotten around to implementing these features yet.
 
+ChangeLog
+---------
+
 A new feature is the lazy opening of datasets. Unlike
 the netCDF interface where attributes for all variables and groups are read at
 the creation of the Dataset object, h5netcdf loads these attributes only
 when accessed.
+
 
 New API
 ~~~~~~~
--- a/README.rst
+++ b/README.rst
@@ -62,6 +62,12 @@ exceptions of:
 
 We simply haven't gotten around to implementing these features yet.
 
+A new feature, enabled by default, is the lazy opening of datasets. Unlike
+the netCDF interface where attributes for all variables and groups are read at
+the creation of the Dataset object, h5netcdf loads these attributes only
+when accessed. Passing `--lazy=False` to `h5netcdf.File` or `h5netcdf.legacyapi.Dataset`
+reverts to the netCDF interface behavior.
+
 New API
 ~~~~~~~
 
--- a/h5netcdf/core.py
+++ b/h5netcdf/core.py
@@ -294,7 +294,7 @@ class Group(Mapping):
                                             **kwargs)
 
         self._variables.set(h5name,self._variable_cls(self, h5name, dimensions))
-        variable=self._variables[name]
+        variable=self._variables[h5name]
 
         if fillvalue is not None:
             value = variable.dtype.type(fillvalue)
--- a/h5netcdf/core.py
+++ b/h5netcdf/core.py
@@ -154,7 +154,7 @@ class lazy_objects(Mapping):
 
     def __getitem__(self, key):
         if not key in self._objects:
-            raise KeyError('Object {0} is not avaialble.'.format(key))
+            raise KeyError(key)
         elif key in self._loaded_objects.keys():
             return self._loaded_objects[key]
         else:
--- a/h5netcdf/core.py
+++ b/h5netcdf/core.py
@@ -132,11 +132,12 @@ class Variable(BaseVariable):
 
 NOT_A_VARIABLE = b'This is a netCDF dimension but not a netCDF variable.'
 
-class lazy_objects(Mapping):
-    def __init__(self, object_cls):
+class _object_names(Mapping):
+    def __init__(self, parent, object_cls):
+        self._parent = parent
+        self._object_cls = object_cls
         self._objects = set()
         self._loaded_objects = dict()
-        self._object_cls = object_cls
 
     def __setitem__(self, name, object):
         self._objects.append(name)
@@ -158,7 +159,7 @@ class lazy_objects(Mapping):
         elif key in self._loaded_objects:
             return self._loaded_objects[key]
         else:
-            self._loaded_objects[key] = self._object_cls(key)
+            self._loaded_objects[key] = self._object_cls(self._parent,key)
             return self[key]
 
 class Group(Mapping):
@@ -178,8 +179,8 @@ class Group(Mapping):
             self._dim_sizes = parent._dim_sizes.new_child()
             self._dim_order = parent._dim_order.new_child()
 
-        self._variables = lazy_objects(lambda x: self._variable_cls(self, x))
-        self._groups = lazy_objects(lambda x: self._group_cls(self, x))
+        self._variables = _object_names(self, self._variable_cls)
+        self._groups = _object_names(self, self._group_cls)
 
         for k, v in self._h5group.items():
             if isinstance(v, h5py.Group):
--- a/h5netcdf/core.py
+++ b/h5netcdf/core.py
@@ -143,6 +143,9 @@ class lazy_objects():
         self._loaded_objects[name]=object
         self._objects[name]=name
 
+    def keys(self):
+        return self._objects.keys()
+
     def __iter__(self):
         for name in self._objects:
             yield name
--- a/h5netcdf/core.py
+++ b/h5netcdf/core.py
@@ -19,8 +19,8 @@ class BaseVariable(object):
 
     def __init__(self, parent, name, dimensions=None):
         self._parent = parent
-        self._root=parent._root
-        self._h5path='/'.join([parent.name.rstrip('/'),name.lstrip('/')])
+        self._root = parent._root
+        self._h5path = '/'.join([parent.name.rstrip('/'), name.lstrip('/')])
         self._dimensions = dimensions
         self._initialized = True
 
@@ -133,15 +133,19 @@ class Variable(BaseVariable):
 NOT_A_VARIABLE = b'This is a netCDF dimension but not a netCDF variable.'
 
 class lazy_objects(Mapping):
-    def __init__(self,object_cls):
-        self._objects=OrderedDict()
-        self._loaded_objects=dict()
-        self._object_cls=object_cls
+    def __init__(self, object_cls):
+        self._objects = list()
+        self._loaded_objects = dict()
+        self._object_cls = object_cls
         return
 
-    def set(self,name,object):
-        self._loaded_objects[name]=object
-        self._objects[name]=name
+    def set(self, name, object):
+        self._loaded_objects[name] = object
+        self._objects.append(name)
+
+    def add(self, name)
+        self._objects.append(name)
+        return
 
     def __iter__(self):
         for name in self._objects:
@@ -150,16 +154,12 @@ class lazy_objects(Mapping):
     def __len__(self):
         return len(self._objects)
 
-    def __setitem__(self,key,value):
-        self._objects[key]=value
-        return
-
-    def __getitem__(self,key):
+    def __getitem__(self, key):
         if key in self._loaded_objects.keys():
             return self._loaded_objects[key]
         else:
-            self._loaded_objects[key]=self._object_cls(self._objects[key])
-            return self[key] 
+            self._loaded_objects[key] = self._object_cls(key)
+            return self[key]
 
 class Group(Mapping):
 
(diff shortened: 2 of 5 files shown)


shoyer:
Thanks for working on this! I'm optimistic this may fix the segmentation fault on Travis-CI that stopped me from merging https://github.com/shoyer/h5netcdf/pull/11.

A few notes:
1. You probably know this already, but distinct changes like are much easier/faster to review in separate PRs.
2. Could you please explain the use case for lazy evaluation for groups and variables? Does this make a meaningful performance difference for you? Are there tradeoffs to adding `lazy=True`? If not, we should switch to that unilaterally.
2. Please add a test for your bug fix -- that's essential to ensuring it remains fixed! See `h5netcdf/tests/test_h5netcdf.py` for examples.

I also have some style suggestions but I'll hold off on them for now.

shoyer:
Please rebase on master to get the fix for your failing test (related to `_NCProperties`) from #19.

laliberte:
Hi Stephan,

Please do suggest styles improvements!

This is my first contribution, so apologies for fitting too much into one pull request.

A short summary of the latest changes:

1) Made `lazy=True` by default. For my use case (large datasets recorded down 10 hierarchical groups), this makes a huge difference. My hunch (based on limited testing) is that smaller datasets would not experience big slow downs.

2) The new lazy implementation that always refer to the h5py.File object does not have a seg fault anymore (`test_reading_str_array_from_netCDF4` with latest Anaconda 2.7 has a seg fault with previous `h5netcdf` version).

3) To ensure a successful build, the tests are now attempting the `char` test using pure `h5py` firstand are performing the test only if it is successful. Here, I was assuming that h5netcdf should fail when h5py fails.

h5netcdf/tests/test_h5netcdf.py
```diff
@@ -158,14 +162,29 @@ def read_legacy_netcdf(tmp_netcdf, read_module, write_module):
...
 
-    v = ds.variables['z']
-    assert array_equal(v, _char_array)
-    assert v.dtype == 'S1'
-    assert v.ndim == 2
-    assert v.dimensions == ('z', 'string3')
-    assert v.ncattrs() == ['_FillValue']
-    assert v.getncattr('_FillValue') == b'X'
+    #Check the behavior if h5py. Cannot expect h5netcdf to overcome these errors:
+    with h5py.File(tmp_netcdf, 'r') as ds:
```
shoyer: Can you move this to a separate utility function, so we can reuse it in both tests? Also please note the underlying bug in netCDF-C in a comment:
https://github.com/Unidata/netcdf-c/issues/298

shoyer:
Is there any reason why one would want to write `lazy=False`? If not, let's remove the option.

Please take a look at [PEP8](https://www.python.org/dev/peps/pep-0008/) for general Python style guidelines. In particular, note the guidelines about whitespace and variable names. I recommend running a PEP8 checker like pyflakes or pep8 to check this automatically while you write your code. pylint is also good, though it's rather strict by default. You can automatically run these tools whenever you save your code.

h5netcdf/core.py
```diff
@@ -125,28 +132,62 @@ def shuffle(self):
...
-        self._variables = OrderedDict()
-        self._groups = OrderedDict()
+        self._variables = lazy_objects(lambda x:self._variable_cls(self,x))
+        self._groups = lazy_objects(lambda x:self._group_cls(self,x))
 
-        for k, v in h5group.items():
+        for k, v in self._h5group.items():
             if isinstance(v, h5py.Group):
-                self._groups[k] = self._group_cls(self, v)
+                self._groups[k] = k
```
shoyer: This feels redundant. I would suggest adding a `add` method to lazy_objects instead.

h5netcdf/core.py
```diff
@@ -125,28 +132,62 @@ def shuffle(self):
 
 NOT_A_VARIABLE = b'This is a netCDF dimension but not a netCDF variable.'
 
+class lazy_objects():
```
shoyer: This should inherit from `collections.Mapping` to ensure it satisfies the mapping interface

h5netcdf/core.py
```diff
@@ -125,28 +132,61 @@ def shuffle(self):
...
-        self._h5group = h5group
+        self._h5path = '/'.join([parent.name.rstrip('/'), name.lstrip('/')])
+
         if parent is not self:
             self._dim_sizes = parent._dim_sizes.new_child()
             self._dim_order = parent._dim_order.new_child()
 
-        self._variables = OrderedDict()
-        self._groups = OrderedDict()
+        self._variables = lazy_objects(lambda x: self._variable_cls(self, x))
```
shoyer: I would change the lazy_objects constructor to take two arguments. It's just a little cleaner than using the lambda.

h5netcdf/core.py
```diff
@@ -125,28 +132,61 @@ def shuffle(self):
 
 NOT_A_VARIABLE = b'This is a netCDF dimension but not a netCDF variable.'
 
+class lazy_objects(Mapping):
+    def __init__(self, object_cls):
+        self._objects = list()
+        self._loaded_objects = dict()
+        self._object_cls = object_cls
+        return
```
shoyer: You don't need all these blank `return` statements. Python does that by default.

h5netcdf/core.py
```diff
@@ -125,28 +132,61 @@ def shuffle(self):
 
 NOT_A_VARIABLE = b'This is a netCDF dimension but not a netCDF variable.'
 
+class lazy_objects(Mapping):
+    def __init__(self, object_cls):
+        self._objects = list()
```
shoyer: I might call this "unloaded_object" for clarity. Also, make it a set so contains checks are fast.

h5netcdf/core.py
```diff
@@ -125,28 +132,61 @@ def shuffle(self):
...
+        self._loaded_objects = dict()
+        self._object_cls = object_cls
+        return
+
+    def set(self, name, object):
```
shoyer: I might just call this `__setitem__` now.

laliberte:
Ok,

1) I've folowed your advice and cleaned up the code using `pylint` and `pyflakes`.
2) I've Made lazy_objects a `collections.Mapping` and added an `add` method. It does not use `OrderedDict` anymore.
3) I've made a more generic function for testing against `h5py`.

h5netcdf/core.py
```diff
@@ -125,28 +132,61 @@ def shuffle(self):
...
+    def __iter__(self):
+        for name in self._objects:
+            yield name
+
+    def __len__(self):
+        return len(self._objects)
+
+    def __getitem__(self, key):
+        if not key in self._objects:
+            raise KeyError('Object {0} is not avaialble.'.format(key))
```
shoyer: `raise KeyError(key)` is consistent with most Python mappings

h5netcdf/core.py
```diff
@@ -125,28 +132,61 @@ def shuffle(self):
...
+        elif key in self._loaded_objects.keys():
```
shoyer: `elif key in self._loaded_objects:` avoids constructing a list for `.keys()` on Python 2.

h5netcdf/tests/test_h5netcdf.py
```diff
@@ -41,6 +41,21 @@ def array_equal(a, b):
 _char_array = string_to_char(np.array(['a', 'b', 'c', 'foo', 'bar', 'baz'],
                                       dtype='S'))
 
+_string_array = np.array([['foobar0', 'foobar1', 'foobar3'],
+                          ['foofoofoo', 'foofoobar', 'foobarbar']])
+
+def is_h5py_char_working(tmp_netcdf, name):
+    #https://github.com/Unidata/netcdf-c/issues/298
```
shoyer: Please add one space after `#` in comments per pep8

README.rst
```diff
@@ -62,6 +62,11 @@ exceptions of:
 
 We simply haven't gotten around to implementing these features yet.
 
+A new feature is the lazy opening of datasets. Unlike
```
shoyer: Could you add a `Changelog` section to the README instead?

laliberte:
Alright, I think I have implemented all your suggestions. Just give me a few minutes and I will push it!

setup.py
```diff
@@ -23,7 +23,7 @@
       long_description=(open('README.rst').read()
                         if os.path.exists('README.rst')
                         else ''),
-      version='0.2.2',
+      version='0.2.3rc1',
```
shoyer: Let's just leave this for now. I'll issue a new release once I test #11

README.rst
shoyer: We might as well call the next version `0.3.0`

h5netcdf/core.py
```diff
@@ -132,21 +132,19 @@ def shuffle(self):
 
 NOT_A_VARIABLE = b'This is a netCDF dimension but not a netCDF variable.'
 
-class lazy_objects(Mapping):
-    def __init__(self, object_cls):
-        self._objects = list()
-        self._loaded_objects = dict()
+class _object_names(Mapping):
```
shoyer: Use `CapWords` for class names per PEP 8. Maybe `_LazyObjectLookup` would be appropriate here?

laliberte:
Sounds good. Let me know if you need anything else.

laliberte:
I like it. I think it's important to have `Lazy` in the name. I think it's a fairly descriptive terminology.

h5netcdf/core.py
```diff
@@ -125,28 +132,59 @@ def shuffle(self):
 
 NOT_A_VARIABLE = b'This is a netCDF dimension but not a netCDF variable.'
 
+class _object_names(Mapping):
+    def __init__(self, parent, object_cls):
+        self._parent = parent
+        self._object_cls = object_cls
+        self._objects = set()
```
shoyer: It occurs to me that `_objects` needs to be ordered, because objects in HDF5/netCDF files have an order which we should iterate over properly.

Python doesn't have an OrderedSet in the standard library, so our best choice might be to use an OrderedDict for `_objects` instead. This suggests an alternative strategy for this class:

- Save unloaded objects in `_objects` as `None`.
- When calling `__getitem__` on an object that is None, replace it with real object.

laliberte: Right, I see. I think that was the reason I switched to `list` but I forgot. Let's use `OrderedDict`, it mimics better HDF5/netCDF.

shoyer:
It might also be a good idea to include a test for object iteration order.

h5netcdf/core.py
```diff
@@ -17,13 +17,20 @@ def _reverse_dict(dict_):
 
 class BaseVariable(object):
 
-    def __init__(self, parent, h5ds, dimensions=None):
+    def __init__(self, parent, name, dimensions=None):
         self._parent = parent
-        self._h5ds = h5ds
+        self._root = parent._root
+        self._h5path = '/'.join([parent.name.rstrip('/'), name.lstrip('/')])
```
shoyer: What is this change about?

h5netcdf/core.py
```diff
@@ -17,13 +17,20 @@ def _reverse_dict(dict_):
...
         self._dimensions = dimensions
         self._initialized = True
 
     @property
+    def _h5ds(self):
+        #Always refer to the root file and store not h5py object
```
shoyer: space after `#`

h5netcdf/core.py
```diff
@@ -162,13 +200,19 @@ def __init__(self, parent, h5group):
...
                     if k.startswith('_nc4_non_coord_'):
-                        name = k[len('_nc4_non_coord_'):]
-                    self._variables[name] = self._variable_cls(self, v)
+                        var_name = k[len('_nc4_non_coord_'):]
+                    self._variables.add(var_name)
         self._initialized = True
 
     @property
+    def _h5group(self):
+        #Always refer to the root file and store not h5py object
```
shoyer: space after `#`

h5netcdf/core.py
laliberte: The `h5py.Dataset` object is no longer passed to the variable class. It is recovered on-demand from the `h5py.File` object on line 28.

h5netcdf/core.py
```diff
@@ -125,28 +132,55 @@ def shuffle(self):
...
+
+    def __iter__(self):
+        for name in self._objects:
+            yield name
+
+    def __len__(self):
+        return len(self._objects)
+
+    def __getitem__(self, key):
+        if self._objects[key] != None:
```
shoyer: It's a best practice to use `is not None` instead of `!= None`: https://www.python.org/dev/peps/pep-0008/#programming-recommendations

h5netcdf/core.py
shoyer: I was wondering particularly about the `rstrip`/`lstrip` part. This seems fine, but let's move this logic (which is repeated twice) to a little utility function, e.g., `_join_h5paths`.

h5netcdf/core.py
```diff
@@ -125,28 +132,55 @@ def shuffle(self):
 
 NOT_A_VARIABLE = b'This is a netCDF dimension but not a netCDF variable.'
 
+class _LazyObjectLookup(Mapping):
+    def __init__(self, parent, object_cls):
+        self._parent = parent
+        self._object_cls = object_cls
+        self._objects = OrderedDict()
+
+    def __setitem__(self, name, object):
```
shoyer: Another minor point: it's considered best practice to avoid shadowing built-ins like `object`. So I would probably better to rename to something else, e..g, `obj`.

laliberte: Done.

h5netcdf/core.py
laliberte: Done

h5netcdf/core.py
laliberte: Done

shoyer:
Thanks @laliberte !

laliberte:
No problems! I might look into adding a few more tweaks to increase compatibility with the netCDF4 API. I've identified a few low hanging fruits already.

I also have been working on a patched `pydap` version that works very well with the ESGF. At the moment I've given it the netCDF4 API but it might make more sense to streamline it with `h5netcdf` since netCDF4 already has DAP support.

shoyer:
> I also have been working on a patched pydap version that works very well with the ESGF. At the moment I've given it the netCDF4 API but it might make more sense to streamline it with h5netcdf since netCDF4 already has DAP support.

Yes, a thin wrapper around scipy, h5netcdf and pydap could be a drop-in replacement for netCDF4-Python (though you could also just use [xarray](http://xarray.pydata.org/en/stable/)). This probably makes more sense as a sister project, though we could also consider renaming h5netcdf.

Issue #21, opened 2016-08-07
Exception AttributeError: "'File' object has no attribute '_closed'" in <object repr() failed> ignore


laliberte:
A `h5netcdf.File` object that failed to open in read-only mode is not properly garbage collected in python 2.7.

```
import h5netcdf
try:
    h5netcdf.File('testfile.nc','r')
except IOError:
    pass

Exception AttributeError: "'File' object has no attribute '_closed'" in <object repr() failed> ignored
```

PR #22, opened 2016-08-07
Fixing the garbage collection issue raised in #21

laliberte:
In python 2.7, `self._closed` has to be set even if `h5py.File` fails to open in read-only mode.

I've created a test for this issue. I was not sure what was the best way to handle this issue. I've gone with a `try` statement. Better suggestions?


h5netcdf/core.py
```diff
@@ -415,13 +415,17 @@ def __repr__(self):
 class File(Group):
 
     def __init__(self, path, mode='a', **kwargs):
-        self._h5file = h5py.File(path, mode, **kwargs)
+        self._closed = True
+        try:
```
shoyer: I think it would be clearer to write:
```python
try:
    self._h5file = ...
except Exception:
    self._closed = True
    raise
else:
    self._closed = False
```
Note that it's important to catch `Exception` rather than use the naked `except`, to avoid catching `KeyboardInterrupt` and the like.

h5netcdf/tests/test_h5netcdf.py
```diff
@@ -482,9 +481,39 @@ def test_reading_str_array_from_netCDF4(tmp_netcdf):
...
+        os.remove(tmp_netcdf)
+    except OSError:
+        pass
+
+def test_failed_read_open_and_clean_delete(tmp_netcdf):
+    # A file that does not exist but is opened for
+    # reading should only raise an IOError and
+    # no AttributeError at garbage collection.
+    try:
+        _silent_remove(tmp_netcdf)
```
shoyer: Rather than removing the tmp file before opening it, I would suggest simply making a path that doesn't exist. This is straightforward in pytest with the [`tmpdir` fixture](http://doc.pytest.org/en/latest/tmpdir.html):
```python
def test_failed_read_open_and_clean_delete(tmpdir):
    path = str(tmpdir.join('this_file_does_not_exist.nc'))
    with h5netcdf.File(path, 'r') as ds:
        ...
```

shoyer:
I think everything will get squashed away in the merge commit, but I see a lot of extra commits in your PR here. Next time, branch off of `master` :).

h5netcdf/tests/test_h5netcdf.py
```diff
@@ -482,9 +481,39 @@ def test_reading_str_array_from_netCDF4(tmp_netcdf):
...
+        with h5netcdf.File(tmp_netcdf, 'r') as ds:
+            pass
+    except IOError:
+        pass
+
+    # Look at garbage collection:
+    import gc
```
shoyer: Imports should go at the top of the module.

h5netcdf/tests/test_h5netcdf.py
```diff
@@ -482,9 +481,39 @@ def test_reading_str_array_from_netCDF4(tmp_netcdf):
...
+    obj_list = gc.get_objects()
```
shoyer: It might be simpler to just write:
```python
gc.collect()  # should not raise
```

h5netcdf/core.py
laliberte: It is clearer. I've changed the code.

h5netcdf/tests/test_h5netcdf.py
laliberte: That's much better, thanks!

h5netcdf/tests/test_h5netcdf.py
laliberte: I've tried different methods and it seems to be the one only that raises an exception within the python execution. Somehow, `gc.collect()` only raises an exception after the python environment in closed and is therefore not picked-up by `pytest`. It's a tough one...

h5netcdf/tests/test_h5netcdf.py
laliberte: Done! About the extra commits, I know. Still getting used to the complete workflow. A few more pull requests and it should be seamless :)

h5netcdf/tests/test_h5netcdf.py
```diff
@@ -482,9 +482,32 @@ def test_reading_str_array_from_netCDF4(tmp_netcdf):
...
+        with h5netcdf.File(path, 'r') as ds:
+            pass
+    except IOError:
+        pass
+
+    # Look at garbage collection:
+    # A simple gc.collect() does not raise an exception.
+    # Must seek the File object and imitate its del command
+    # by forcing it to close.
+    obj_list = gc.get_objects()
```
shoyer: I wonder if simply calling `ds.close()` would trigger this, without needing to get the garbage collector involved.

laliberte: I wondered the same and it raises a `NameError`. The current method to raise a runtime exception is the only one I could find. It's a fairly tricky one to force to surface...

h5netcdf/tests/test_h5netcdf.py
shoyer: Maybe even just writing `ds.close()` here would do the trick?

h5netcdf/tests/test_h5netcdf.py
shoyer: OK, fair enough :).

I tried:
```
In [2]: try:
   ...:     f = h5netcdf.File('not-a-real-file.nc', 'r')
   ...: finally:
   ...:     f.close()
```
but I also just get `NameError: name 'f' is not defined`.

shoyer:
Thanks again @laliberte. Let me know when you're done with PRs for now and I'll issue v0.3.1.

PR #24, opened 2016-09-01
Fix missing attribute exception when writing to groups

bnaul:
See: pydata/xarray#994.

I also made a small change in `core.Group._create_dimension` when checking for unlimited dimensions: presently this throws an exception when `size=0`, which seems incorrect.

### DIFF
--- a/h5netcdf/core.py
+++ b/h5netcdf/core.py
@@ -379,6 +379,10 @@ class Group(Mapping):
     def parent(self):
         return self._parent
 
+    def flush(self):
+        self._root.flush()
+    sync = flush
+
     @property
     def groups(self):
         return Frozen(self._groups)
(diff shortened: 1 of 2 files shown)


shoyer:
Thanks!

These are simple enough that I'm pretty they are correct, but if there's a straightforward way to write tests that would be even better...

bnaul:
I added a test for a writing a length-0 dimension, but I'm not sure what an appropriate test would be for the `sync/flush` error; the problem was `xarray` was calling that method explicitly and it didn't exist. Maybe there should be a test in `xarray.backends.H5NetCDFStore`?

shoyer:
> Maybe there should be a test for xarray.backends.H5NetCDFStore?

Yeah, that's probably the most straightforward place to check this.

We could also add a sync test in h5netcdf, just to verify that it doesn't error.

bnaul:
Added a `sync` call to the `h5netcdf` tests, hopefully it's good to go now!

shoyer:
Awesome, thanks @bnaul !

I will go ahead an issue an new release shortly

shoyer:
v0.3.1 is on pypi

PR #26, opened 2017-02-26
Include tests to the release tarball

ghisvail:
The current tarball hosted on PyPI is missing the test suite associated with the package. I added it manually to the source of the [Debian package](https://packages.debian.org/source/sid/python-h5netcdf), but it would be better if they were systematically included in future releases on PyPI, hence this PR.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,1 +1,2 @@
 include LICENSE
+recursive-include *.py


ghisvail:
`h5netcdf` has been accepted to [Debian](https://packages.debian.org/source/sid/python-h5netcdf). However, I had to patch the code base to include the tests, which is not ideal.

Could you please consider merging this proposal to include the tests in `sdist` and uploading a new tarball to PyPI (`v0.3.2`?).

Thanks.

shoyer:
Thanks!

I can issue a new release if someone else is interested in this but otherwise I am inclined to wait for now given that you already got it into Debian.

ghisvail:
Thanks for merging this. I'd like to keep the amount of patching applied to the upstream source code to a minimum, hence myself asking for the new release. I am happy to wait if you have further work scheduled on the package.

PR #29, opened 2017-08-29
Warnings when using invalid netCDF features

### LINKED ISSUE #28: Disable writing files with features not supported by netCDF-C?
Currently, we support several features implicitly via h5py that make for files the netCDF-C library can't read:

- complex dtypes (https://github.com/pydata/xarray/issues/748)
- multidimensional attribute arrays (see http://stackoverflow.com/questions/42963789)
- probably some other dtypes that h5py handles

I am inclined to disable writing files with these features to avoid incompatibilities with standard netCDF tools. We would need to add some sort of encoding for handling complex data directly with xarray first to support that common use case.

shoyer:
Fixes #28
xref pydata/xarray#1536

New behavior:
```
In [1]: import h5netcdf

In [2]: f = h5netcdf.File('warning.h5')

In [3]: f.attrs['foo'] = 1j
/Users/shoyer/conda/envs/h5netcdf-dev/bin/ipython:1: FutureWarning: complex dtypes
are supported by h5py, but not part of the NetCDF API. You are writing an HDF5
file that is not a valid NetCDF file! In the future, this will be an error, unless
you set invalid_netcdf=True.
  #!/bin/bash /Users/shoyer/conda/envs/h5netcdf-dev/bin/python.app

In [4]: f = h5netcdf.File('error.h5', invalid_netcdf=False)

In [5]: f.attrs['foo'] = 1j
CompatibilityError: complex dtypes are not a supported NetCDF feature, and are not
allowed by h5netcdf unless invalid_netcdf=True.
```

CC @dopplershift -- I would appreciate any feedback you might have on the error/warning message and the new README.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ matrix:
   - python: 2.7
   - python: 3.4
   - python: 3.5
+  - python: 3.6
   - python: 2.7
     env: H5PY_VERSION="=2.1"
   allow_failures:
--- a/README.rst
+++ b/README.rst
@@ -45,11 +45,11 @@ Usage
 -----
 
 h5netcdf has two APIs, a new API and a legacy API. Both interfaces currently
-reproduce most of the features of the netCDF interface, with the noteable
+reproduce most of the features of the netCDF interface, with the notable
 exceptions of:
 
 - support for operations the rename or delete existing objects.
-- suport for creating unlimited dimensions.
+- support for creating unlimited dimensions.
 
 We simply haven't gotten around to implementing these features yet. Patches
 would be very welcome.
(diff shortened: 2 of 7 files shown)


dopplershift:
I think the language for both is strong enough to convey clearly to users what they're doing. I also like the fact the "private" netCDF attributes are disabled when using unsupported features. Thanks @shoyer !

shoyer:
> I also like the fact the "private" netCDF attributes are disabled when using unsupported features.

Just to be clear, I don't write `_NCProperties` (which holds provenance information for netCDF files). I do still write the private `_Netcdf4Coordinates` and `_Netcdf4Dimid` when necessary: these are part of the data model.

dopplershift:
Sure, and that makes sense.

PR #30, opened 2017-09-05
Fixup inclusion of the tests in the source tarball

ghisvail:
My mistake for not verifying my first patch before hand. The tarball for version 0.4.0 on PyPI still lacks the test suite. This commit fixes it.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,2 +1,2 @@
 include LICENSE
-recursive-include *.py
+recursive-include h5netcdf/tests *.py


shoyer:
oops -- I should have tested this, too!

Thanks

ghisvail:
Could you publish a hotfix release for this (0.4.1 perhaps?) so I can update the Debian packaging with the tests? Thanks.

shoyer:
Done

PR #33, opened 2017-10-05
Adding support for unlimited dimensions

krischer:
This PR adds support for unlimited dimensions.

I'm not very familiar with netncdf but as far as I can tell the underlying HDF5 data sets are automatically resized if required so that is what I implemented here. I think I caught the most common indexing patterns.

It is tested in the sense that netcdf can open the resulting files just fine and everything appears to be correct. There is also full test coverage for all changes and this PR does not change the behavior of limited dimensions.

It would be great if this could be merged with a subsequent release fairly soon as we have a number of downstream project that depend on it and we really need unlimited dimensions. Let me know if I can help with things!

### DIFF
--- a/h5netcdf/core.py
+++ b/h5netcdf/core.py
@@ -215,7 +215,7 @@ class Group(Mapping):
         # Flag used to determine if the current size of the unlimited
         # dimensions must be detected manually at the end of the
         # initialization.
-        _current_size_set = False
+        has_existing_unlimited_dims = False
 
         for k, v in self._h5group.items():
             if isinstance(v, h5py.Group):
@@ -236,7 +236,7 @@ class Group(Mapping):
 
                     self._dim_sizes[k] = size
                     self._current_dim_sizes[k] = current_size
-                    _current_size_set = True
+                    has_existing_unlimited_dims = True
                     if dim_id is None:
                         dim_id = len(self._dim_order)
                     self._dim_order[k] = dim_id
@@ -246,7 +246,7 @@ class Group(Mapping):
                         var_name = k[len('_nc4_non_coord_'):]
                     self._variables.add(var_name)
 
-        if _current_size_set is True:
+        if has_existing_unlimited_dims is True:
             self._determine_current_dimension_sizes()
 
         self._initialized = True
--- a/h5netcdf/core.py
+++ b/h5netcdf/core.py
@@ -122,21 +122,6 @@ class BaseVariable(object):
     def __setitem__(self, key, value):
         self._h5ds[key] = value
 
-    def resize(self, size, axis=None):
-        """
-        Change the shape of a dataset.
-
-        This is just a thin wrapper around h5py's ``Dataset.resize()`` method.
-
-        ``size`` may be a tuple giving the new dataset shape, or an integer
-        giving the new length of the specified axis.
-
-        Datasets may be resized only up to ``Dataset.maxshape``. In netCDF
-        semantics this means that it can only be resized along unlimited
-        dimensions.
-        """
-        self._h5ds.resize(size=size, axis=axis)
-
     @property
     def attrs(self):
         return Attributes(self._h5ds.attrs,
@@ -221,6 +206,7 @@ class Group(Mapping):
 
         if parent is not self:
             self._dim_sizes = parent._dim_sizes.new_child()
+            self._current_dim_sizes = parent._current_dim_sizes.new_child()
             self._dim_order = parent._dim_order.new_child()
 
         self._variables = _LazyObjectLookup(self, self._variable_cls)
@@ -236,14 +222,18 @@ class Group(Mapping):
                         assert dim_id is not None
                         coord_ids = v.attrs['_Netcdf4Coordinates']
                         size = v.shape[list(coord_ids).index(dim_id)]
+                        current_size = size
                     else:
                         assert len(v.shape) == 1
                         # Unlimited dimensions are represented as None.
                         if v.maxshape == (None,):
                             size = None
+                            current_size = v.size
                         else:
                             size = v.size
+                            current_size = v.size
                     self._dim_sizes[k] = size
+                    self._current_dim_sizes[k] = current_size
                     if dim_id is None:
                         dim_id = len(self._dim_order)
                     self._dim_order[k] = dim_id
@@ -269,6 +259,7 @@ class Group(Mapping):
             raise ValueError('dimension %r already exists' % name)
 
         self._dim_sizes[name] = size
+        self._current_dim_sizes[name] = 0 if size is None else size
         self._dim_order[name] = len(self._dim_order)
 
     @property
@@ -348,23 +339,18 @@ class Group(Mapping):
                                     file=self._root,
                                     stacklevel=stacklevel)
 
-        shape = tuple(self.dimensions[d] for d in dimensions)
         if name in self.dimensions and name not in dimensions:
             h5name = '_nc4_non_coord_' + name
         else:
             h5name = name
 
-        # None in the shape indicates unlimited dimensions. Set the maxshape
-        # to the shape in that case and replace all None's with zero for the
-        # actual shape. Or if there is data replace by the actual data size.
-        # This will allow resizing the data sets later if required.
-        if None in shape:
-            kwargs['maxshape'] = shape
-            if data is None:
-                shape = tuple(i if i is not None else 0 for i in shape)
-            else:
-                shape = tuple(j if j is not None else i
-                              for i, j in zip(data.shape, shape))
+        shape = tuple(self._current_dim_sizes[d] for d in dimensions)
+        maxshape = tuple(self._dim_sizes[d] for d in dimensions)
+
+        # If it is passed directly it will change the default compression
+        # settings.
+        if shape != maxshape:
+            kwargs["maxshape"] = maxshape
 
         self._h5group.create_dataset(h5name, shape, dtype=dtype,
                                      data=data, fillvalue=fillvalue,
@@ -434,6 +420,10 @@ class Group(Mapping):
                 dims = self._variables[dim].dimensions
                 coord_ids = np.array([dim_order[d] for d in dims], 'int32')
                 h5ds.attrs['_Netcdf4Coordinates'] = coord_ids
+            # Might have to be resized if it already exists.
+            elif h5ds.maxshape == (None, ) and \
+                    self._current_dim_sizes[dim] != h5ds.shape[0]:
+                h5ds.resize((self._current_dim_sizes[dim],))
 
             scale_name = dim if dim in self.variables else NOT_A_VARIABLE
             h5ds.dims.create_scale(h5ds, scale_name)
@@ -496,6 +486,30 @@ class Group(Mapping):
                   % (self._cls_name, self.name, len(self)))
         return '\n'.join([header] + self._repr_body())
 
+    def resize_dimension(self, dimension, size):
+        """
+        Resize a dimension to a certain size.
+
+        This will pad with zeros where necessary.
+        """
+        if self.dimensions[dimension] is not None:
+            raise ValueError("Only unlimited dimensions can be resized.")
+
+        # Resize the dimension.
+        self._current_dim_sizes[dimension] = size
+        self._dim_order[dimension] = len(self._dim_order)
+
+        for var in self.variables.values():
+            try:
+                idx = var.dimensions.index(dimension)
+            except ValueError:
+                continue
+            var._h5ds.resize(size, axis=idx)
+
+        # Recurse as dimensions are visible to this group and all child groups.
+        for i in self.groups:
+            i.resize(dimension, size)
+
 
 class File(Group):
 
@@ -508,8 +522,14 @@ class File(Group):
             raise
         else:
             self._closed = False
+
+        # Three maps to keep track of dimensions in terms of size (might be
+        # unlimited), current size (identical to size for limited dimensions),
+        # and their position.
         self._dim_sizes = ChainMap()
+        self._current_dim_sizes = ChainMap()
         self._dim_order = ChainMap()
+
         self._mode = mode
         self._root = self
         self._h5path = '/'
--- a/h5netcdf/core.py
+++ b/h5netcdf/core.py
@@ -500,15 +500,17 @@ class Group(Mapping):
         self._dim_order[dimension] = len(self._dim_order)
 
         for var in self.variables.values():
-            try:
-                idx = var.dimensions.index(dimension)
-            except ValueError:
-                continue
-            var._h5ds.resize(size, axis=idx)
+            new_shape = list(var.shape)
+            for _i, d in enumerate(var.dimensions):
+                if d == dimension:
+                    new_shape[_i] = size
+            new_shape = tuple(new_shape)
+            if new_shape != var.shape:
+                var._h5ds.resize(new_shape)
 
         # Recurse as dimensions are visible to this group and all child groups.
-        for i in self.groups:
-            i.resize(dimension, size)
+        for i in self.groups.values():
+            i.resize_dimension(dimension, size)
 
 
 class File(Group):
--- a/h5netcdf/core.py
+++ b/h5netcdf/core.py
@@ -226,12 +226,8 @@ class Group(Mapping):
                     else:
                         assert len(v.shape) == 1
                         # Unlimited dimensions are represented as None.
-                        if v.maxshape == (None,):
-                            size = None
-                            current_size = v.size
-                        else:
-                            size = v.size
-                            current_size = v.size
+                        size = None if v.maxshape == (None,) else v.size
+                        current_size = v.size
 
                     self._dim_sizes[k] = size
                     self._current_dim_sizes[k] = current_size
@@ -428,12 +424,10 @@ class Group(Mapping):
         dim_order = self._dim_order.maps[0]
         for dim in sorted(dim_order, key=lambda d: dim_order[d]):
             if dim not in self._h5group:
-                size = self.dimensions[dim]
-                if size is None:
-                    size = 0
-                    kwargs = {"maxshape": (None, )}
-                else:
-                    kwargs = {}
+                size = self._current_dim_sizes[dim]
+                kwargs = {}
+                if self._dim_sizes[dim] is None:
+                    kwargs["maxshape"] = (None,)
                 self._h5group.create_dataset(
                     name=dim, shape=(size,), dtype='S1', **kwargs)
 
@@ -444,10 +438,6 @@ class Group(Mapping):
                 dims = self._variables[dim].dimensions
                 coord_ids = np.array([dim_order[d] for d in dims], 'int32')
                 h5ds.attrs['_Netcdf4Coordinates'] = coord_ids
-            # Might have to be resized if it already exists.
-            elif h5ds.maxshape == (None, ) and \
-                    self._current_dim_sizes[dim] != h5ds.shape[0]:
-                h5ds.resize((self._current_dim_sizes[dim],))
 
             scale_name = dim if dim in self.variables else NOT_A_VARIABLE
             h5ds.dims.create_scale(h5ds, scale_name)
@@ -495,10 +485,9 @@ class Group(Mapping):
     def _repr_body(self):
         return (
             ['Dimensions:'] +
-            ['    %s: %s' % (k, ("Unlimited (current: %s)" %
-                                 self._current_dim_sizes[k])
-                             if v is None else v)
-             for k, v in self.dimensions.items()] +
+            ['    %s: %s' % (
+             k, ("Unlimited (current: %s)" % self._current_dim_sizes[k])
+             if v is None else v) for k, v in self.dimensions.items()] +
             ['Groups:'] +
             ['    %s' % g for g in self.groups] +
             ['Variables:'] +
@@ -521,17 +510,17 @@ class Group(Mapping):
         This will pad with zeros where necessary.
         """
         if self.dimensions[dimension] is not None:
-            raise ValueError("Only unlimited dimensions can be resized.")
+            raise ValueError("Dimension '%s' is not unlimited and thus "
+                             "cannot be resized." % dimension)
 
         # Resize the dimension.
         self._current_dim_sizes[dimension] = size
-        self._dim_order[dimension] = len(self._dim_order)
 
         for var in self.variables.values():
             new_shape = list(var.shape)
-            for _i, d in enumerate(var.dimensions):
+            for i, d in enumerate(var.dimensions):
                 if d == dimension:
-                    new_shape[_i] = size
+                    new_shape[i] = size
             new_shape = tuple(new_shape)
             if new_shape != var.shape:
                 var._h5ds.resize(new_shape)
--- a/h5netcdf/core.py
+++ b/h5netcdf/core.py
@@ -469,15 +469,19 @@ class Group(Mapping):
     _cls_name = 'h5netcdf.Group'
 
     def _repr_body(self):
-        return (['Dimensions:'] +
-                ['    %s: %s' % (k, v) for k, v in self.dimensions.items()] +
-                ['Groups:'] +
-                ['    %s' % g for g in self.groups] +
-                ['Variables:'] +
-                ['    %s: %r %s' % (k, v.dimensions, v.dtype)
-                 for k, v in self.variables.items()] +
-                ['Attributes:'] +
-                ['    %s: %r' % (k, v) for k, v in self.attrs.items()])
+        return (
+            ['Dimensions:'] +
+            ['    %s: %s' % (k, ("Unlimited (current: %s)" %
+                                 self._current_dim_sizes[k])
+                             if v is None else v)
+             for k, v in self.dimensions.items()] +
+            ['Groups:'] +
+            ['    %s' % g for g in self.groups] +
+            ['Variables:'] +
+            ['    %s: %r %s' % (k, v.dimensions, v.dtype)
+             for k, v in self.variables.items()] +
+            ['Attributes:'] +
+            ['    %s: %r' % (k, v) for k, v in self.attrs.items()])
 
     def __repr__(self):
         if self._root._closed:
--- a/h5netcdf/core.py
+++ b/h5netcdf/core.py
@@ -30,6 +30,12 @@ def _join_h5paths(parent_path, child_path):
     return '/'.join([parent_path.rstrip('/'), child_path.lstrip('/')])
 
 
+def _name_from_dimension(dim):
+    # First value in a dimension is the actual dimension scale
+    # which we'll use to extract the name.
+    return dim[0].name.split('/')[-1]
+
+
 class CompatibilityError(Exception):
     """Raised when using features that are not part of the NetCDF4 API."""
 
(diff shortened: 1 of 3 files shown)


h5netcdf/core.py
```diff
@@ -121,6 +120,72 @@ def __getitem__(self, key):
         return self._h5ds[key]
 
     def __setitem__(self, key, value):
+        # This conditional is used to automatically resize the underlying HDF5
+        # data set if the corresponding dimension is unlimited and if the value
+        # requires it.
+        if self._h5ds.maxshape and None in self._h5ds.maxshape:
```
shoyer: Can you move this logic for inferring the new size for a variable into a separate helper function? That would make things a little easier to test.

h5netcdf/core.py
```diff
@@ -121,6 +120,72 @@ def __getitem__(self, key):
         return self._h5ds[key]
 
     def __setitem__(self, key, value):
+        # This conditional is used to automatically resize the underlying HDF5
```
shoyer: Rather than automatically resizing if the `value` is too large, I would prefer that the core h5netcdf API match h5py:
- Add a `resize` method that calls the h5py method of the same name.
- Only put automatic resizing in the `netCDF4` compatibility layer.

We could potential expose the `needs_resize()` helper function as part of the public API if you're interested in using it with the non-legacy PAI.

h5netcdf/core.py
```diff
@@ -121,6 +120,72 @@ def __getitem__(self, key):
...
+                                 "the variable.")
+            # First step is to figure out the required shape of the data set
+            # so that the value can fit.
+            #
+            # Special case handling for setting the full data set.
+            if key == slice(None) or key is Ellipsis:
+                minimum_required_shape = value.shape
+            else:
+                index = [slice(None) for _ in range(self._h5ds.ndim)]
+                for _i, _j in enumerate(
```
shoyer: style note: just use normal variable names for arguments inside a `for` loop.

krischer:
Sorry for the delay, especially because I was asking for handling this quickly...

I changed it to only offer an explicit `resize()` method using the new API and will put the automatic resizing logic into a separate PR. Let me know if this still needs some work.

h5netcdf/core.py
```diff
@@ -338,6 +354,18 @@ def _create_child_variable(self, name, dimensions, dtype, data, fillvalue,
...
             h5name = name
 
+        # None in the shape indicates unlimited dimensions. Set the maxshape
+        # to the shape in that case and replace all None's with zero for the
+        # actual shape. Or if there is data replace by the actual data size.
+        # This will allow resizing the data sets later if required.
+        if None in shape:
+            kwargs['maxshape'] = shape
+            if data is None:
+                shape = tuple(_i if _i is not None else 0 for _i in shape)
```
shoyer: Can you please drop the `_` prefix here? I understand the value of not leaking the variable into the outer scope, but this is harder to read and not a problem on Python 3, anyways.

h5netcdf/core.py
```diff
@@ -338,6 +354,18 @@ def _create_child_variable(self, name, dimensions, dtype, data, fillvalue,
...
+            else:
+                shape = tuple(_j if _j is not None else data.shape[_i]
+                              for _i, _j in enumerate(shape))
```
shoyer: Use `zip(data.shape, shape)` rather than `enumerate()`

h5netcdf/tests/test_h5netcdf.py
```diff
@@ -570,3 +568,41 @@ def test_invalid_then_valid_no_ncproperties(tmp_netcdf):
...
+    # Close and read again to also test correct parsing of unlimited
+    # dimensions.
+    with h5netcdf.File(tmp_netcdf) as f:
+        assert f.dimensions['x'] is None
+        assert f._h5file['x'].maxshape == (None,)
+        assert f._h5file['x'].shape == (0,)
+
+def test_writing_to_an_unlimited_dimension(tmp_netcdf):
+    with h5netcdf.File(tmp_netcdf) as f:
+        # Two dimension, only one is unlimited.
```
shoyer: dimensions

h5netcdf/core.py
```diff
@@ -123,6 +122,21 @@ def __getitem__(self, key):
     def __setitem__(self, key, value):
         self._h5ds[key] = value
 
+    def resize(self, size, axis=None):
```
shoyer: Thinking about this a little more, I'm concerned that this could allow for creating netCDF files with inconsistent dimension sizes between variables. This shouldn't be possible with h5netcdf's public API.

Can we move this to a `Group` level instead, and resize all variables contained inside the group that use the dimension at once? The signature would look something like `group.resize(size, dimension)` where `dimension` is the name of the dimension to be resized.

h5netcdf/tests/test_h5netcdf.py
```diff
@@ -570,3 +568,41 @@ def test_invalid_then_valid_no_ncproperties(tmp_netcdf):
...
+        # Set the whole array.
+        f.variables['dummy2'].resize((1, 3))
+        f.variables['dummy2'][:] = [[1, 2, 3]]
+        np.testing.assert_allclose(f.variables['dummy2'], [[1, 2, 3]])
+
+        # With an ellipsis
+        f.variables['dummy2'].resize((2, 3))
+        f.variables['dummy2'][...] = [[1, 2, 3], [4, 5, 6]]
+        np.testing.assert_allclose(f.variables['dummy2'],
+                                   [[1, 2, 3], [4, 5, 6]])
```
shoyer: I'm a little concerned about test coverage here. Even if we don't support it with our wrapper yet, I'd like automated tests that verify that we can read files with unlimited dimensions written by netCDF4 and write files readable by netCDF4.

h5netcdf/core.py
krischer: Sure.

h5netcdf/core.py
krischer: Sure. It just a habit I picked up due to scoping issues. It is still an problem with for loop indices in Python 3 but I agree that it is less serious nowadays.

h5netcdf/tests/test_h5netcdf.py
krischer: I'll add some files generated with the C bindings.

krischer: I just realized that netCDF4 is already a test dependency - so I'll just generate them on the fly.

krischer:
Okay - I had to dig a bit deeper than expected and it is getting more complicated. There is now a `resize_dimension()` method at the group level and it tries to assure that all variables with that dimension are always consistent. Please have a look at the tests which should explain the behavior.

Files generated by `netCDF4` can be read by `h5netcdf`, and vice versa.

Note that the always consistent file is not exactly the behavior of `netCDF` - it does not appear to keep a file internally consistent and I'm not sure why.

Another thing that is kind of awkward: `h5netcdf` at init time (and for unlimited dimensions) will look at all variables with that dimension to determine the dimension's current size. I'm not sure how netcdf does this but I could not find anything in the hdf5 files it generates (see the currently last test).

I'm also not very familiar with netcdf semantics so it would be great if you can have a look.

I can work a bit more on this tomorrow if necessary.

h5netcdf/core.py
```diff
@@ -222,10 +222,19 @@ def __init__(self, parent, name):
...
                     else:
                         assert len(v.shape) == 1
-                        size = v.size
+                        # Unlimited dimensions are represented as None.
+                        if v.maxshape == (None,):
+                            size = None
+                            current_size = v.size
+                        else:
+                            size = v.size
+                            current_size = v.size
```
shoyer: This is the same in both conditions, so it doesn't need to be in the condition. You might also be able to fit the condition on one line, e.g.,
```python
size = None if v.maxshape == (None,) else v.size
current_size = v.size
```

h5netcdf/core.py
```diff
@@ -445,15 +493,19 @@ def attrs(self):
...
-                ['Variables:'] +
-                ['    %s: %r %s' % (k, v.dimensions, v.dtype)
-                 for k, v in self.variables.items()] +
-                ['Attributes:'] +
-                ['    %s: %r' % (k, v) for k, v in self.attrs.items()])
+        return (
+            ['Dimensions:'] +
+            ['    %s: %s' % (k, ("Unlimited (current: %s)" %
+                                 self._current_dim_sizes[k])
+                             if v is None else v)
```
shoyer: indent three spaces to match `(`, not `k`

h5netcdf/core.py
```diff
@@ -234,6 +243,29 @@ def __init__(self, parent, name):
                     if k.startswith('_nc4_non_coord_'):
                         var_name = k[len('_nc4_non_coord_'):]
                     self._variables.add(var_name)
+
+        # One last pass to get the current sizes.
+        #
+        # I'm not entirely sure how the netcdf C api determines the current
+        # size - should probably be looked into.
```
shoyer: I'm actually not sure we need or want section. As long as resizing is done consistently across all variables (necessary to ensure a valid netCDF file), just looking at the dimension scale should be enough.

h5netcdf/core.py
```diff
@@ -462,6 +514,32 @@ def __repr__(self):
...
         return '\n'.join([header] + self._repr_body())
 
+    def resize_dimension(self, dimension, size):
+        """
+        Resize a dimension to a certain size.
+
+        This will pad with zeros where necessary.
+        """
+        if self.dimensions[dimension] is not None:
+            raise ValueError("Only unlimited dimensions can be resized.")
```
shoyer: please print the name of the offending dimension

h5netcdf/core.py
```diff
@@ -462,6 +514,32 @@ def __repr__(self):
...
+
+        # Resize the dimension.
+        self._current_dim_sizes[dimension] = size
+        self._dim_order[dimension] = len(self._dim_order)
+
+        for var in self.variables.values():
+            new_shape = list(var.shape)
+            for _i, d in enumerate(var.dimensions):
```
shoyer: please remove prefacing underscore :)

h5netcdf/core.py
```diff
@@ -400,6 +444,10 @@ def _create_dim_scales(self):
                 dims = self._variables[dim].dimensions
                 coord_ids = np.array([dim_order[d] for d in dims], 'int32')
                 h5ds.attrs['_Netcdf4Coordinates'] = coord_ids
+            # Might have to be resized if it already exists.
+            elif h5ds.maxshape == (None, ) and \
```
shoyer: I'm confused by the current placement of this clause. I think it belongs about 6 lines up, after `if dim not in self._h5group`.

shoyer: Also, use parentheses to span multiple lines instead of explicit continuations, e.g.,
```
elif (h5ds.maxshape == (None, ) and
      self._current_dim_sizes[dim] != h5ds.shape[0]):
```

h5netcdf/tests/test_h5netcdf.py
```diff
@@ -570,3 +573,166 @@ def test_invalid_then_valid_no_ncproperties(tmp_netcdf):
...
+        np.testing.assert_allclose(f.variables['dummy2'],
+                                   [[1, 2, 3], [5, 6, 7]])
+
+        f.variables['dummy3'][...] = [[1, 2, 3], [5, 6, 7]]
+        np.testing.assert_allclose(f.variables['dummy3'],
+                                   [[1, 2, 3], [5, 6, 7]])
+
+def test_c_api_can_read_unlimited_dimensions(tmp_netcdf):
+    with h5netcdf.File(tmp_netcdf) as f:
+        # Two dimensions, only one is unlimited.
```
shoyer: Three dimensions now

h5netcdf/core.py
```diff
@@ -400,6 +444,10 @@ def _create_dim_scales(self):
                 dims = self._variables[dim].dimensions
                 coord_ids = np.array([dim_order[d] for d in dims], 'int32')
                 h5ds.attrs['_Netcdf4Coordinates'] = coord_ids
+            # Might have to be resized if it already exists.
```
shoyer: How would this situation come up? My current thinking is that either:

- The variable corresponding to the dimension scale already existed. If any resizing was done, the variable also got resized via `resize_dimension()`.
- If the dimension scale didn't exist, it's created here with the `create_dataset`.

h5netcdf/core.py
```diff
@@ -462,6 +514,32 @@ def __repr__(self):
...
+        Resize a dimension to a certain size.
+
+        This will pad with zeros where necessary.
+        """
+        if self.dimensions[dimension] is not None:
+            raise ValueError("Only unlimited dimensions can be resized.")
+
+        # Resize the dimension.
+        self._current_dim_sizes[dimension] = size
+        self._dim_order[dimension] = len(self._dim_order)
```
shoyer: Why does `dim_order` get changed when you resize?

h5netcdf/tests/test_h5netcdf.py
```diff
@@ -570,3 +573,166 @@ def test_invalid_then_valid_no_ncproperties(tmp_netcdf):
...
+
+        # This is parsed correctly due to h5netcdf's init trickery.
+        assert f._current_dim_sizes['x'] == 2
+        assert f._current_dim_sizes['y'] == 3
+        assert f._current_dim_sizes['z'] == 0
+
+        # But the actual data-set and arrays are not correct.
+        assert f['dummy1'].shape == (2, 3)
+        # XXX: This array has some data with dimension x - netcdf does not
+        # appear to keep dimensions consistent.
```
shoyer: Hmm. It might be worth filing a bug report.

krischer:
Regarding: https://github.com/shoyer/h5netcdf/pull/33#discussion_r143879258

I'll comment here cause it will otherwise get hidden once I push to this branch and I think its important.

> I'm actually not sure we need or want section. As long as resizing is done consistently across all variables (necessary to ensure a valid netCDF file), just looking at the dimension scale should be enough.

This will work if the file has been written with h5netcdf, but files written with netCDF4 don't show this behavior. And even if this is bug within netCDF4 (as you mention below) h5netcdf would somehow have to work around it to be able to work with existing netcdf files. And h5netcdf does need access to the current size of an unlimited dimension to be able to properly work with it.

The following example does not change the dimension scale when data is assigned to it. And in the `h5dump` output I don't see any way to get the current size of a dimension aside from looking at all datasets.

But I'll remove the current loop and will use the `REFERENCE_LIST`  of a dimensions to get its data sets - should be quite a bit faster especially with nested files.

```python
import netCDF4

with netCDF4.Dataset("out.nc", "w") as f:
    f.createDimension('x', None)
    dummy1 = f.createVariable('dummy1', float, ('x'))
    dummy1[:] = [1, 2, 3, 4]
```

```bash
$ h5dump out.nc
HDF5 "out.nc" {
GROUP "/" {
   ATTRIBUTE "_NCProperties" {
      DATATYPE  H5T_STRING {
         STRSIZE 57;
         STRPAD H5T_STR_NULLTERM;
         CSET H5T_CSET_ASCII;
         CTYPE H5T_C_S1;
      }
      DATASPACE  SCALAR
      DATA {
      (0): "version=1|netcdflibversion=4.4.1.1|hdf5libversion=1.8.18"
      }
   }
   DATASET "dummy1" {
      DATATYPE  H5T_IEEE_F64LE
      DATASPACE  SIMPLE { ( 4 ) / ( H5S_UNLIMITED ) }
      DATA {
      (0): 1, 2, 3, 4
      }
      ATTRIBUTE "DIMENSION_LIST" {
         DATATYPE  H5T_VLEN { H5T_REFERENCE { H5T_STD_REF_OBJECT }}
         DATASPACE  SIMPLE { ( 1 ) / ( 1 ) }
         DATA {
         (0): (DATASET 331 /x )
         }
      }
   }
   DATASET "x" {
      DATATYPE  H5T_IEEE_F32BE
      DATASPACE  SIMPLE { ( 0 ) / ( H5S_UNLIMITED ) }
      DATA {
      }
      ATTRIBUTE "CLASS" {
         DATATYPE  H5T_STRING {
            STRSIZE 16;
            STRPAD H5T_STR_NULLTERM;
            CSET H5T_CSET_ASCII;
            CTYPE H5T_C_S1;
         }
         DATASPACE  SCALAR
         DATA {
         (0): "DIMENSION_SCALE"
         }
      }
      ATTRIBUTE "NAME" {
         DATATYPE  H5T_STRING {
            STRSIZE 64;
            STRPAD H5T_STR_NULLTERM;
            CSET H5T_CSET_ASCII;
            CTYPE H5T_C_S1;
         }
         DATASPACE  SCALAR
         DATA {
         (0): "This is a netCDF dimension but not a netCDF variable.         0"
         }
      }
      ATTRIBUTE "REFERENCE_LIST" {
         DATATYPE  H5T_COMPOUND {
            H5T_REFERENCE { H5T_STD_REF_OBJECT } "dataset";
            H5T_STD_I32LE "dimension";
         }
         DATASPACE  SIMPLE { ( 1 ) / ( 1 ) }
         DATA {
         (0): {
               DATASET 628 /dummy1 ,
               0
            }
         }
      }
   }
}
}
```

h5netcdf/tests/test_h5netcdf.py
krischer: Might do once this PR is done. I'm not sure if this is a bug or intended behavior. I spent a couple minutes reading the netCDF spec but could not really find how it is supposed to behave. I unfortunately don't have the time right now to read it in detail.

h5netcdf/core.py
krischer: 🙈

h5netcdf/core.py
krischer: Yea this is a mistake.

h5netcdf/core.py
krischer: You are right - a remnant from the old logic - the data sets can be created with the correct size right away.

h5netcdf/core.py
krischer: This is the only indentation `flake8` accepts and in most parts of your code you adhere to it. I'll break it in a different way while still being `flake8` compliant so it looks a bit nicer.

krischer:
I implemented all the review points and rebased on the latest mater. Let me know how you want the consistency when handling unlimited dimensions to be handled.

krischer:
Just a quick ping in case you forgot. Sorry for being a bit pesky!

Also please let me know if you think this will still take a while - then I'll find some internal work-around for our dependent packages.

h5netcdf/core.py
```diff
@@ -234,6 +239,46 @@ def __init__(self, parent, name):
...
+        # One last pass to get the current sizes. This is necessary as h5netcdf
+        # needs to now the current size of each unlimited dimension.
+        #
+        # I'm not entirely sure how the netcdf C api determines the current
+        # size - should probably be looked into.
+        def _find_dim(h5group, dim):
+            if dim not in h5group:
+                # Might happen if the dimension has not been created yet.
+                if h5group.name == "/":
+                    return None
```
shoyer: I don't think this can happen here. We are only checking dimensions in `self.dimensions`, which looks like at `self._dim_sizes` -- which in turn is only filled in where there is already a matching dimension scale.

It would be better to raise an error here (maybe `AssertionError`?) because this really shouldn't ever happen.

h5netcdf/core.py
```diff
@@ -234,6 +239,46 @@ def __init__(self, parent, name):
...
+                return _find_dim(h5group.parent, dim)
+            return h5group[dim]
+
+        for d in self.dimensions:
+            if self.dimensions[d] is not None:
+                continue
+            dim = _find_dim(self._h5group, d)
+            if not dim:
+                continue
```
shoyer: As noted above, I think we can drop this.

h5netcdf/core.py
```diff
@@ -234,6 +239,46 @@ def __init__(self, parent, name):
...
+
+            # Assume the reference list at each dimension is set correctly.
+            if "REFERENCE_LIST" not in dim.attrs:
+                continue
```
shoyer: If something ended up in `self.dimensions`, it's a dimension scale *somewhere* in the group or a parent group. So we should be able to find that dimensions scale in `_find_dim()`. Possibly we need to add the check for "REFERENCE_LIST" up above into `_find_dim()`.

And if this can't happen, we shouldn't have this `continue` here. It's good to flexibly accommodate malformed input files, but we should not allow internal consistency (in h5netcdf) to fail -- that is the way of bugs.

h5netcdf/core.py
```diff
@@ -234,6 +239,46 @@ def __init__(self, parent, name):
...
+
+        # One last pass to get the current sizes. This is necessary as h5netcdf
+        # needs to now the current size of each unlimited dimension.
+        #
+        # I'm not entirely sure how the netcdf C api determines the current
+        # size - should probably be looked into.
+        def _find_dim(h5group, dim):
+            if dim not in h5group:
+                # Might happen if the dimension has not been created yet.
+                if h5group.name == "/":
```
shoyer: I don't think this can happen here. We are only checking dimensions in `self.dimensions`, which looks like at `self._dim_sizes` -- which in turn is only filled in where there is already a matching dimension scale.

h5netcdf/tests/test_h5netcdf.py
```diff
@@ -571,3 +574,167 @@ def test_invalid_then_valid_no_ncproperties(tmp_netcdf):
...
+
+def test_writing_to_an_unlimited_dimension(tmp_netcdf):
+    with h5netcdf.File(tmp_netcdf) as f:
+        # Two dimensions, only one is unlimited.
+        f.dimensions['x'] = None
+        f.dimensions['y'] = 3
+
+        # Cannot create it without first resizing it.
+        with pytest.raises(ValueError) as e:
+                f.create_variable('dummy1', data=np.array([[1, 2, 3]]),
```
shoyer: dedent this block.

h5netcdf/tests/test_h5netcdf.py
```diff
@@ -571,3 +574,167 @@ def test_invalid_then_valid_no_ncproperties(tmp_netcdf):
...
+                                  dimensions=('x', 'y'))
+                assert e.value.args[0] == \
```
shoyer: Use parentheses for line continuation instead of explicit `\`:
https://www.python.org/dev/peps/pep-0008/#maximum-line-length

> The preferred way of wrapping long lines is by using Python's implied line continuation inside parentheses, brackets and braces. Long lines can be broken over multiple lines by wrapping expressions in parentheses. These should be used in preference to using a backslash for line continuation.

h5netcdf/core.py
krischer: With the lazy evaluation it can happen - all my changed lines (except the `#pragma: no cover` ones) are covered by the tests. But we can skip the whole dimension detection logic if the init actually created the file because then we are already internally consistent. I'll change it.

h5netcdf/core.py
krischer: Same comment as above.

h5netcdf/core.py
```diff
@@ -462,6 +520,32 @@ def __repr__(self):
                   % (self._cls_name, self.name, len(self)))
         return '\n'.join([header] + self._repr_body())
 
+    def resize_dimension(self, dimension, size):
+        """
+        Resize a dimension to a certain size.
+
+        This will pad with zeros where necessary.
```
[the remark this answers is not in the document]
krischer: True - I was not aware data sets have a  `fillvalue` attribute. Does NetCDF support setting that attribute? The NetCDF fill value appears to be something different: http://www.unidata.ucar.edu/software/netcdf/docs/fill_values.html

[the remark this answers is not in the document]
shoyer: At least for netCDF4/HDF5, I'm pretty sure there's a there's a distinction between `_FillValue` (the attribute) and `fillvalue` (the HDF5 property). In any case, it's definitely not always zero.

h5netcdf/core.py
shoyer: Lazy evaluation controls creating sub-groups/child-variables in h5netcdf, but it shouldn't dictate what appears in `self._h5group` (which is based on what's in the file per h5py). So I'm somewhat mystified how this can happen.

Can you point to which test triggers this line (e.g., if you add a print statement)?

krischer: It can no longer happen with my latest changes as the logic to determine the current sizes of the unlimited dimensions is now only called when reading from an existing file.

It might be that I misunderstood how the lazy evaluation actually works but the actual dimension scales in the HDF5 file are only created during the `flush()` call, so initializing a new file, adding a dimension, and then creating a new group would call the `__init__()` method with `self.dimensions` being set, without dimension scales being in the file (yet).

krischer:
Ready for another review round.

h5netcdf/core.py
```diff
@@ -234,8 +245,52 @@ def __init__(self, parent, name):
...
+        """
+        def _find_dim(h5group, dim):
+            if dim not in h5group:
+                return _find_dim(h5group.parent, dim)
+            return h5group[dim]
+
+        for d in self.dimensions:
+            if self.dimensions[d] is not None:
+                continue
+            dim = _find_dim(self._h5group, d)
```
shoyer: minor nit: more descriptive variables names for `dim` (`dim_variable`?) and `d` (`dim_name`?) would be appreciated

h5netcdf/core.py
shoyer: OK, thanks for your patience here. You're right -- we have two forms of laziness here (and I had forgotten how my code worked!).

h5netcdf/core.py
```diff
@@ -234,8 +245,52 @@ def __init__(self, parent, name):
...
+        """
+        Helper method to determine the current size of all unlimited
+        dimensions.
+        """
+        def _find_dim(h5group, dim):
+            if dim not in h5group:
+                return _find_dim(h5group.parent, dim)
+            return h5group[dim]
+
+        for d in self.dimensions:
```
shoyer: Would it make sense to only iterate over `self._dim_sizes.maps[-1]` here instead of `self.dimensions`?

This would only pick up dimensions newly newly defined in this group (in the file), not a parent group, so it would avoid some redundant work.

And if the file has new dimensions and is not yet flushed to disk, those new dimensions wouldn't be iterated over, either. So we wouldn't need the boolean `has_existing_unlimited_dims`.

krischer: That is not possible as more then one dimension might be set during the init. I found a different way though: The current size is now determined per dimension and only when it is necessary at the time the dimension is set.

shoyer: (Correction, I was actually thinking of `self._dim_sizes.maps[0]`.)

That should pull out the mapping with everything set in the most recent "layer" of the ChainMap, which means everything set in this `Group`: https://docs.python.org/3/library/collections.html#collections.ChainMap

I'll take a look at your solution.

shoyer:
Big thanks @krischer! I'll issue a new release shortly, probably later today.

krischer:
Great :) Thanks for your help!

I still have the automatic resizing logic flying around in some branch but I'm not sure it is necessary especially as the logic for it is really fairly complex. Raise your hand if you still want it and I'll open a separate PR with the current state.

shoyer:
v0.5.0 is out now on pypi and conda-forge.

krischer:
Thanks a lot for ! Also for the very useful package!

In case you are looking for another success story of `h5netcdf`: We utilize it in [pyexodus](https://github.com/SalvusHub/pyexodus) to build exodus files which are meshes we use for large-scale high order finite element wave propagation simulations (see http://salvus.io).

The bottom line is that it enables us to skip two large C codes (NetCDF + Exodus) and it ends up being up to 100 times faster and more memory efficient.

PR #41, opened 2017-12-15
Add h5pyd

ajelenak-thg:
This PR would enable h5netcdf users to access HDF5 data server via the HDF5 REST API if the h5pyd package is installed.

### DIFF
--- a/h5netcdf/core.py
+++ b/h5netcdf/core.py
@@ -562,11 +562,16 @@ class File(Group):
 
     def __init__(self, path, mode='a', invalid_netcdf=None, **kwargs):
         try:
-            if path.startswith('http'):
+            if path.startswith(('http', 'hdf5:')):
                 if no_h5pyd:
                     raise RuntimeError('h5pyd package required')
+                try:
+                    with h5pyd.File(path, 'r') as f:  # noqa
+                        pass
+                    self._preexisting_file = True
+                except OSError:
+                    self._preexisting_file = False
                 self._h5file = h5pyd.File(path, mode, **kwargs)
-                self._preexisting_file = True if 'r' in mode else False
             else:
                 self._preexisting_file = os.path.exists(path)
                 self._h5file = h5py.File(path, mode, **kwargs)
--- a/h5netcdf/core.py
+++ b/h5netcdf/core.py
@@ -16,8 +16,10 @@ try:
     import h5pyd
 except ImportError:
     no_h5pyd = True
+    h5_group_types = (h5py.Group,)
 else:
     no_h5pyd = False
+    h5_group_types = (h5py.Group, h5pyd.Group)
 
 __version__ = '0.5.1'
 
@@ -225,8 +227,7 @@ class Group(Mapping):
         self._groups = _LazyObjectLookup(self, self._group_cls)
 
         for k, v in self._h5group.items():
-            if ((not no_h5pyd and isinstance(v, h5pyd.Group)) or
-                    isinstance(v, h5py.Group)):
+            if isinstance(v, h5_group_types):
                 # add to the groups collection if this is a h5py(d) Group
                 # instance
                 self._groups.add(k)
@@ -562,9 +563,11 @@ class File(Group):
 
     def __init__(self, path, mode='a', invalid_netcdf=None, **kwargs):
         try:
-            if path.startswith(('http', 'hdf5:')):
+            if path.startswith(('http://', 'https://', 'hdf5://')):
                 if no_h5pyd:
-                    raise RuntimeError('h5pyd package required')
+                    raise ImportError(
+                        "No module named 'h5pyd'. h5pyd is required for "
+                        "opening remote paths with h5netcdf: {}".format(path))
                 try:
                     with h5pyd.File(path, 'r') as f:  # noqa
                         pass
--- a/h5netcdf/core.py
+++ b/h5netcdf/core.py
@@ -12,6 +12,12 @@ from .attrs import Attributes
 from .dimensions import Dimensions
 from .utils import Frozen
 
+try:
+    import h5pyd
+except ImportError:
+    no_h5pyd = True
+else:
+    no_h5pyd = False
 
 __version__ = '0.5.0'
 
@@ -560,9 +566,15 @@ class Group(Mapping):
 class File(Group):
 
     def __init__(self, path, mode='a', invalid_netcdf=None, **kwargs):
-        self._preexisting_file = os.path.exists(path)
         try:
-            self._h5file = h5py.File(path, mode, **kwargs)
+            if path.startswith('http'):
+                if no_h5pyd:
+                    raise RuntimeError('h5pyd package required')
+                self._h5file = h5pyd.File(path, mode, **kwargs)
+                self._preexisting_file = True if 'r' in mode else False
+            else:
+                self._preexisting_file = os.path.exists(path)
+                self._h5file = h5py.File(path, mode, **kwargs)
         except Exception:
             self._closed = True
             raise
--- a/h5netcdf/core.py
+++ b/h5netcdf/core.py
@@ -572,7 +572,7 @@ class File(Group):
                     with h5pyd.File(path, 'r') as f:  # noqa
                         pass
                     self._preexisting_file = True
-                except (OSError, IOError):
+                except IOError:
                     self._preexisting_file = False
                 self._h5file = h5pyd.File(path, mode, **kwargs)
             else:
--- a/h5netcdf/core.py
+++ b/h5netcdf/core.py
@@ -569,7 +569,7 @@ class File(Group):
                     with h5pyd.File(path, 'r') as f:  # noqa
                         pass
                     self._preexisting_file = True
-                except OSError:
+                except (OSError, IOError):
                     self._preexisting_file = False
                 self._h5file = h5pyd.File(path, mode, **kwargs)
             else:
--- a/h5netcdf/core.py
+++ b/h5netcdf/core.py
@@ -225,7 +225,10 @@ class Group(Mapping):
         self._groups = _LazyObjectLookup(self, self._group_cls)
 
         for k, v in self._h5group.items():
-            if isinstance(v, h5py.Group):
+            if ((not no_h5pyd and isinstance(v, h5pyd.Group)) or
+                    isinstance(v, h5py.Group)):
+                # add to the groups collection if this is a h5py(d) Group
+                # instance
                 self._groups.add(k)
             else:
                 if v.attrs.get('CLASS') == b'DIMENSION_SCALE':
--- a/h5netcdf/tests/test_h5netcdf.py
+++ b/h5netcdf/tests/test_h5netcdf.py
@@ -49,6 +49,7 @@ _char_array = string_to_char(np.array(['a', 'b', 'c', 'foo', 'bar', 'baz'],
 _string_array = np.array([['foobar0', 'foobar1', 'foobar3'],
                           ['foofoofoo', 'foofoobar', 'foobarbar']])
 
+
 def is_h5py_char_working(tmp_netcdf, name):
     # https://github.com/Unidata/netcdf-c/issues/298
     with h5py.File(tmp_netcdf, 'r') as ds:
@@ -61,6 +62,7 @@ def is_h5py_char_working(tmp_netcdf, name):
                 return False
             else:
                 raise
+
 
 def write_legacy_netcdf(tmp_netcdf, write_module):
     ds = write_module.Dataset(tmp_netcdf, 'w')
@@ -194,7 +196,8 @@ def read_legacy_netcdf(tmp_netcdf, read_module, write_module):
                            'shuffle': False, 'zlib': False}
     ds.close()
 
-    #Check the behavior if h5py. Cannot expect h5netcdf to overcome these errors:
+    # Check the behavior if h5py. Cannot expect h5netcdf to overcome these
+    # errors:
     if is_h5py_char_working(tmp_netcdf, 'z'):
         ds = read_module.Dataset(tmp_netcdf, 'r')
         v = ds.variables['z']
@@ -248,7 +251,8 @@ def read_h5netcdf(tmp_netcdf, write_module):
     if not PY2 and write_module is not netCDF4:
         # skip for now: https://github.com/Unidata/netcdf4-python/issues/388
         assert ds.attrs['other_attr'] == 'yes'
-    assert set(ds.dimensions) == set(['x', 'y', 'z', 'empty', 'string3', 'mismatched_dim'])
+    assert set(ds.dimensions) == set(['x', 'y', 'z', 'empty', 'string3',
+                                      'mismatched_dim'])
     assert set(ds.variables) == set(['foo', 'y', 'z', 'intscalar', 'scalar',
                                      'var_len_str', 'mismatched_dim'])
     assert set(ds.groups) == set(['subgroup'])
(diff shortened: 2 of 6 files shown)


shoyer:
It would be good to add a basic integration test to ensure that this actually works (and doesn't break in the future).

Does h5pyd have any facilitates for setting up a local test server?

rsignell-usgs:
@ajelenak-thg , just curious if you are planning on adding the basic test.

ajelenak-thg:
What would that basic test do? For a URL-looking path:

* if h5pyd is not installed verify that an exception is being raised;
* If h5pyd is installed test... (what?)

There is no yet an HDF5 REST server that would be always accessible. One possibility is to set up an [h5serv](https://github.com/HDFGroup/h5serv) prior to running h5netcdf tests. Is this acceptable?

shoyer:
> One possibility is to set up an h5serv prior to running h5netcdf tests. Is this acceptable?

Yes, sounds like a good option for an integration test. Ideally the setup/tear-down would be done from Python functions (rather than separate shell scripts), so the test-suite still works with pytest assuming h5server/h5pyd are installed:
https://docs.pytest.org/en/2.7.3/xunit_setup.html

jreadey:
@ajelenak-thg - for h5pyd (https://github.com/HDFGroup/h5pyd) I use the HDF Cloud beta instance as an endpoint with the username/password setup as a Travis setting.

How about this approach for h5netcdf testing?  We can ask @shoyer to add the Travis environment variables.

ajelenak-thg:
@jreadey That is certainly an option. Alternatively, we could just test that:

1. If h5pyd is installed, an exception is raised because there is no HDF5 REST server to respond.
2. If h5pyd is not installed, an exception is raised about requiring h5pyd.

These two cases cover the new code that is being added.

shoyer:
A simple unit test to verify that `h5pyd.File` gets called (using mock) would have some value.

An external test hitting an HDF Cloud server would be even better -- that could verify that things actually work. In my experience, the additional point of failure with a dependency on external services can be annoying, but it's still much better than nothing.

Ideally we would have an automated integration test somewhere that approximates h5netcdf's existing test suite, but routes through a local server running h5serv/h5pyd. That's what I would need to be confident that everything in h5netcdf is working correctly.

ajelenak-thg:
These two tests check the code added to h5netcdf to route URL-like paths to h5pyd.
```python
def test_mocked_remote_open_with_h5pyd():
    path = 'http://test.url'
    with mock.patch.object(h5netcdf.Group, '__init__', lambda a, b, c: None), \
            mock.patch('h5netcdf.core.no_h5pyd', False), \
            mock.patch('h5pyd.File') as mock_h5pyd_open:
        h5netcdf.File(path, 'r')
        mock_h5pyd_open.assert_called_with(path, 'r')


def test_mocked_remote_open_without_h5pyd_raises():
    with mock.patch('h5netcdf.core.no_h5pyd', True):
        with pytest.raises(RuntimeError):
            h5netcdf.File('http://test.url', 'r')
```
Using `mock` for the `h5netcdf.Group` class avoids the file content scan and also makes the tests independent from h5pyd availability. The `RuntimeError` exception comes from the new code in h5netcdf when h5pyd is not installed but a URL-like path is used. I can change the type of exception if another is deemed more appropriate.

> Another good option, which we are currently using with pydap in the xarray test suite, is to construct a "remote" File directly using internal h5pyd routines instead of accessing a remote server. Does h5pyd have any public APIs suitable for creating a simple File directly?

What do you mean by "directly"? h5pyd must have a live HDF5 REST server to function. The PyDAP tests in xarray seem to compare a local netCDF file with its copy on the test OPeNDAP server.

If the goal is to test h5netcdf with h5pyd in addition to h5py then we need to have an HDF5 server somewhere. An always-on test server does not (yet) exist like in the OPeNDAP's case. My colleague @jreadey suggested this in a previous [comment](https://github.com/shoyer/h5netcdf/pull/41#issuecomment-356616755) on this PR.

shoyer:
> What do you mean by "directly"? h5pyd must have a live HDF5 REST server to function. The PyDAP tests in xarray seem to compare a local netCDF file with its copy on the test OPeNDAP server.

PyDAP also has an [internal data model](http://www.pydap.org/en/latest/developer.html#the-dap-data-model) that allows datasets to be constructed directly. We use that in xarray so that some of our tests don't rely on an external server:
https://github.com/pydata/xarray/blob/0a0593d78fad6c0b776d4c3c6b32a24b2bdfba35/xarray/tests/test_backends.py#L2040-L2054

ajelenak-thg:
No such capability in h5pyd. The DAP model sits above storage data models like HDF5 or netCDF Common Data Model. h5pyd is similar to h5py; any storage operation is immediately sent to the HDF5 server.

So what now?

jreadey:
@ajelenak-thg - how about adding this functionality to h5pyd?  i.e. if endpoint is None, or the path is relative, have h5pyd.File invoke h5py.File.

shoyer:
> how about adding this functionality to h5pyd? i.e. if endpoint is None, or the path is relative, have h5pyd.File invoke h5py.File.

I'm not sure that would really help for the goal of an h5netcdf/h5pyd integration test :).

It seems like we would ideally be testing h5nertcdf/h5pyd/h5serv all together, but that might be a bit tricky to setup for tests on Travis-CI?

jreadey:
@shoyer - not that hard to test with the remote server on Travis.  This is what I have for the h5pyd Travis test: https://travis-ci.org/HDFGroup/h5pyd.
You'll note that it test using three different drivers:

- h5py with local files
- h5pyd with local h5serv
- h5pyd with remote hsds

Issue with the last one is that if the server is down or there is a random internet flake out, you'll get spurious errors.  In practice it hasn't been a problem with the h5pyd Travis test.  I think there was one time I had to restart a failed Travis job.

ajelenak-thg:
And the issue with the second one (h5serv) is that it lacks support for the HDF5 dimension scale, which is critical for netCDF interoperability. @jreadey, is this correct?

jreadey:
Yes that is correct.  Unless we go back and add dimension scale support to h5serv, only hsds would be relevant.

shoyer:
> Issue with the last one is that if the server is down or there is a random internet flake out, you'll get spurious errors. In practice it hasn't been a problem with the h5pyd Travis test. I think there was one time I had to restart a failed Travis job.

OK, I would be happy with that. h5netcdf isn't like xarray -- the pace of pull requests is slow enough that I can deal with the occasional flaky test. Is there a remote hsds server we can point to for this?

ajelenak-thg:
Excellent! We now have a target... There is a server used by beta testers and for our own dev work. The Travis configuration will need a couple of additional environment variables.

ajelenak-thg:
Never say never!

@shoyer please take a look at the test code and let me know if its on the right track. The goal was to run the tests with both h5py/netcdf4 and h5pyd. Two tests currently fail because of the same h5pyd issue (support for the `complex128` datatype). Once that is fixed this PR should be ready for merge.

The main changes:

1. The `tmp_netcdf()` fixture generates both local and remote HDF5 temporary files. The remote file tests will be skipped by this fixture if any of the following conditions is satisfied:

    * `pytest` command is not invoked with the `--restapi` option,
    * h5pyd is not available (not importable),
    * either of the `$HS_USERNAME` or `$HS_PASSWORD` environment variables is not defined.

1. All tests involving the netCDF4 package are skipped for remote HDF5 files because the HDF5 library does not yet support HDF5 REST API.

1. In the tests that previously used h5py there is now a new helper function `get_hdf5_module()` that returns either h5py or h5pyd depending on the `tmp_netcdf()`'s value.

1. `.travis.yml` has the required information about a test HDF5 REST server. If that server is alive when Travis runs tests the `--restapi` command option will be included in the `pytest` command. Environment variables `$HS_USERNAME` and `$HS_PASSWORD`, which provide test user credentials, are not in this file and have to be defined in the [Repository Settings](https://docs.travis-ci.com/user/environment-variables/#Defining-Variables-in-Repository-Settings).

shoyer:
Do I understand correctly that you're using h5pyd for both reading *and* writing files?

ajelenak-thg:
Yes.

h5netcdf/tests/test_h5netcdf.py
```diff
@@ -11,10 +14,35 @@
...
 
+try:
+    import h5pyd
+    without_h5pyd = False
+except ImportError:
+    without_h5pyd = True
+
+
+@pytest.fixture(params=['testfile.nc', 'hdf5://testfile'])
+def tmp_netcdf(request, tmpdir):
```
shoyer: Rather than adding lots of `pytest.skip` invocations below (which is a little ugly), I would suggest writing two different fixtures:
- `tmp_local_netcdf`: like the original version of `tmp_netcdf`
- `tmp_local_or_remote_netcdf`: parameterized to either be a local or remote URL (i.e., this version of `tmp_netcdf`)

Then each test can simply use the appropriate fixture.

h5netcdf/core.py
```diff
@@ -219,7 +225,10 @@ def __init__(self, parent, name):
         self._groups = _LazyObjectLookup(self, self._group_cls)
 
         for k, v in self._h5group.items():
-            if isinstance(v, h5py.Group):
+            if ((not no_h5pyd and isinstance(v, h5pyd.Group)) or
+                    isinstance(v, h5py.Group)):
```
shoyer: I think a cleaner way to write this would make a top level constant:
```
if no_h5pyd:
    h5_group_types = (h5py.Group,)
else:
    h5_group_types = (h5py.Group, h5pyd.Group)
```

Then use simply `isinstance(v, h5_group_types)`

h5netcdf/core.py
```diff
@@ -552,9 +561,20 @@ def resize_dimension(self, dimension, size):
 class File(Group):
 
     def __init__(self, path, mode='a', invalid_netcdf=None, **kwargs):
-        self._preexisting_file = os.path.exists(path)
         try:
-            self._h5file = h5py.File(path, mode, **kwargs)
+            if path.startswith(('http', 'hdf5:')):
+                if no_h5pyd:
+                    raise RuntimeError('h5pyd package required')
```
shoyer: I think the cleanest exception type here would be `IOError`.

Also, let's make the error message more informative:
```
raise ImportError("No module named 'h5pyd'. h5pyd is required for "
                  "opening remote paths with h5netcdf: {}".format(path))
```

h5netcdf/core.py
```diff
@@ -552,9 +561,20 @@ def resize_dimension(self, dimension, size):
...
+                try:
+                    with h5pyd.File(path, 'r') as f:  # noqa
+                        pass
+                    self._preexisting_file = True
+                except (OSError, IOError):
```
shoyer: Can h5pyd really raise both these errors if a file isn't found? I would think one would be enough...

h5netcdf/core.py
```diff
@@ -552,9 +561,20 @@ def resize_dimension(self, dimension, size):
 class File(Group):
 
     def __init__(self, path, mode='a', invalid_netcdf=None, **kwargs):
-        self._preexisting_file = os.path.exists(path)
         try:
-            self._h5file = h5py.File(path, mode, **kwargs)
+            if path.startswith(('http', 'hdf5:')):
```
shoyer: nit: I would prefer being even more explicit here with `("http://", "https://", "hdf5://")` to match the checks in h5pyd:
https://github.com/HDFGroup/h5pyd/blob/741a8e4abacdfc0df15d9d045db0bd2cd6f47d94/h5pyd/_hl/files.py#L129

h5netcdf/tests/test_h5netcdf.py
```diff
@@ -11,10 +14,35 @@
...
+    if request.param.startswith(('http:', 'hdf5:')):
```
shoyer: please put these prefixes into a module level constant that you can re-use throughout

shoyer:
> Yes

OK, very cool!

> Environment variables $HS_USERNAME and $HS_PASSWORD, which provide test user credentials, are not in this file and have to be defined in the Repository Settings.

Sounds good, please email these to me (shoyer@gmail.com) and I will add them into Travis.

h5netcdf/core.py
ajelenak-thg: This issue came up when testing in Travis with Python 2.7. For Pythons 3: `OSError`; for Python 2.7: `IOError`.

shoyer: OK, please add a brief comment here to explain.

Looking at the [Python docs](https://docs.python.org/3/library/exceptions.html#IOError):

> The following exceptions are kept for compatibility with previous versions; starting from Python 3.3, they are aliases of OSError.
>
> exception EnvironmentError
> exception IOError
> exception WindowsError
> Only available on Windows.

So I think it would probably be OK to use `IOError` for both cases.

h5netcdf/tests/test_h5netcdf.py
```diff
@@ -268,7 +276,7 @@ def read_legacy_netcdf(tmp_netcdf, read_module, write_module):
 
 
 def read_h5netcdf(tmp_netcdf, write_module):
-    remote_h5 = True if tmp_netcdf.startswith(('http:', 'hdf5:')) else False
+    remote_h5 = True if tmp_netcdf.startswith(REMOTE_H5_PATH_PREFIX) else False
```
shoyer: should just be: `remote_h5 = tmp_netcdf.startswith(REMOTE_H5_PATH_PREFIX)`

ajelenak-thg: Done.

shoyer:
Can you try pushing another commit so can verify that the h5pyd tests run on Travis? I believe I've added the environment variables.

Maybe update the README to briefly mention h5pyd, and add a brief note to the changelog. (My plan is to issue the 0.6 release immediately after merging this PR.)

shoyer:
I still don't see the env variables showing up in the build.

Have I done something wrong in my Travis-CI config?

image

ajelenak-thg:
The Travis config looks good. That is how I set them up, too.

Not an expert on this but perhaps the env. variables are not being set for you because this is a branch in my fork?

shoyer:
It appears that this is an intentional design decision by Travis:

"Similarly, we do not provide these values to untrusted builds, triggered by pull requests from another repository."
https://docs.travis-ci.com/user/environment-variables/#Defining-Variables-in-Repository-Settings

So I guess we should make sure Travis-CI passes on your fork before merging this. Based on https://travis-ci.org/ajelenak-thg/h5netcdf/builds/387952771 it looks like there are two tests to fix:
h5netcdf/tests/test_h5netcdf.py::test_invalid_netcdf_warns[hdf5://testfile] FAILED [ 62%]
h5netcdf/tests/test_h5netcdf.py::test_invalid_netcdf_okay[hdf5://testfile] FAILED [ 72%]

ajelenak-thg:
> Based on https://travis-ci.org/ajelenak-thg/h5netcdf/builds/387952771 it looks like there are two tests to fix:
> h5netcdf/tests/test_h5netcdf.py::test_invalid_netcdf_warns[hdf5://testfile] FAILED [ 62%]
> h5netcdf/tests/test_h5netcdf.py::test_invalid_netcdf_okay[hdf5://testfile] FAILED [ 72%]

Correct. Both are because h5pyd does not support `dtype('complex128')` yet. This should get fixed soon.

shoyer:
Okay, in that case let's skip or xfail those cases for now, and we can
merge this!

ajelenak-thg:
I skipped two two tests for remote files and updated the README.

shoyer:
Thanks @ajelenak-thg !

Since h5pyd doesn't require h5py, we might want to consider making h5py an optional requirement at some point in the future.

PR #51, opened 2019-02-20
enable loading h5py file like objects

scottyhq:
Goes along will pull request in xarray - https://github.com/pydata/xarray/pull/2782

 note that commenting the pre-existing file check was a quick change to see if this works. The path checking logic will need to be changed before merging.

### DIFF
--- a/h5netcdf/core.py
+++ b/h5netcdf/core.py
@@ -577,7 +577,7 @@ class File(Group):
 
     def __init__(self, path, mode='a', invalid_netcdf=None, **kwargs):
         try:
-            if path.startswith(('http://', 'https://', 'hdf5://')):
+            if isinstance(path, str) and path.startswith(('http://', 'https://', 'hdf5://')):
                 if no_h5pyd:
                     raise ImportError(
                         "No module named 'h5pyd'. h5pyd is required for "
@@ -590,7 +590,7 @@ class File(Group):
                     self._preexisting_file = False
                 self._h5file = h5pyd.File(path, mode, **kwargs)
             else:
-                self._preexisting_file = os.path.exists(path)
+                #self._preexisting_file = os.path.exists(path)
                 self._h5file = h5py.File(path, mode, **kwargs)
         except Exception:
             self._closed = True
--- a/h5netcdf/core.py
+++ b/h5netcdf/core.py
@@ -7,6 +7,8 @@ import warnings
 import h5py
 import numpy as np
 
+from distutils.version import LooseVersion
+
 from .compat import ChainMap, OrderedDict, unicode
 from .attrs import Attributes
 from .dimensions import Dimensions
@@ -594,12 +596,12 @@ class File(Group):
                     self._preexisting_file = os.path.exists(path)
                     self._h5file = h5py.File(path, mode, **kwargs)
             else:  # file-like object
-                if h5py.__version__ < '2.9.0':
+                if h5py.__version__ < LooseVersion('2.9.0'):
                     raise ImportError(
-                        "h5py version ({}) must be greater than 2.9.0 to load \
-                        file-like objects.".format(h5py.__version__))
+                        "h5py version ({}) must be greater than 2.9.0 to load "
+                        "file-like objects.".format(h5py.__version__))
                 else:
-                    self._preexisting_file = False
+                    self._preexisting_file = mode in {'r', 'r+', 'a'}
                     self._h5file = h5py.File(path, mode, **kwargs)
         except Exception:
             self._closed = True
--- a/h5netcdf/core.py
+++ b/h5netcdf/core.py
@@ -593,8 +593,8 @@ class File(Group):
                 else:
                     self._preexisting_file = os.path.exists(path)
                     self._h5file = h5py.File(path, mode, **kwargs)
-            else: #file-like object
-                if h5py.__version__ >= '2.9.0':
+            else:  # file-like object
+                if h5py.__version__ < '2.9.0':
                     raise ImportError(
                         "h5py version ({}) must be greater than 2.9.0 to load \
                         file-like objects.".format(h5py.__version__))
--- a/h5netcdf/core.py
+++ b/h5netcdf/core.py
@@ -577,21 +577,30 @@ class File(Group):
 
     def __init__(self, path, mode='a', invalid_netcdf=None, **kwargs):
         try:
-            if isinstance(path, str) and path.startswith(('http://', 'https://', 'hdf5://')):
-                if no_h5pyd:
+            if isinstance(path, str):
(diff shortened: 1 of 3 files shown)


shoyer:
Very nice!

It would be great to have a test that verifies that it works to pass in a file-like object (if using a . sufficiently new version of h5py).

h5netcdf/core.py
```diff
@@ -577,21 +577,30 @@ class File(Group):
...
+                            pass
+                        self._preexisting_file = True
+                    except IOError:
+                        self._preexisting_file = False
+                    self._h5file = h5pyd.File(path, mode, **kwargs)
+                else:
+                    self._preexisting_file = os.path.exists(path)
+                    self._h5file = h5py.File(path, mode, **kwargs)
+            else:  # file-like object
+                if h5py.__version__ < '2.9.0':
```
shoyer: Prefer to use `distutils.version.LooseVersion` for comparing version numbers, e.g., `h5py.__version__< distutils.version.LooseVersion('2.9.0')`. This version would break if h5py ever releases a version 2.10.0.

h5netcdf/core.py
```diff
@@ -577,21 +577,30 @@ class File(Group):
...
                     raise ImportError(
-                        "No module named 'h5pyd'. h5pyd is required for "
-                        "opening remote paths with h5netcdf: {}".format(path))
-                try:
-                    with h5pyd.File(path, 'r') as f:  # noqa
-                        pass
-                    self._preexisting_file = True
-                except IOError:
+                        "h5py version ({}) must be greater than 2.9.0 to load \
+                        file-like objects.".format(h5py.__version__))
```
shoyer: Use implicit string concatenation here instead of backslash:
```
                        "h5py version ({}) must be greater than 2.9.0 to load "
                        "file-like objects.".format(h5py.__version__))
```

h5netcdf/core.py
```diff
@@ -577,21 +577,30 @@ class File(Group):
...
+                else:
                     self._preexisting_file = False
```
shoyer: Can we check the `mode` argument here instead?

I think we want `self._preexisting_file = mode in {'r', 'r+', 'a'}` (we would rather be conservative about assuming that a file *may* exist, because otherwise we write extra metadata to the file indicating that it's definitely a netCDF file).
http://docs.h5py.org/en/stable/high/file.html#opening-creating-files

h5netcdf/tests/test_h5netcdf.py
```diff
@@ -412,6 +417,16 @@ def test_write_legacyapi_read_h5netcdf(tmp_local_netcdf):
     read_h5netcdf(tmp_local_netcdf, legacyapi)
 
 
+def test_fileobj():
+    # requires h5py>2.9.0
```
shoyer: use `pytest.skipif` to skip this if the right version of h5py is not installed

h5netcdf/tests/test_h5netcdf.py
```diff
@@ -48,10 +50,10 @@ def tmp_local_or_remote_netcdf(request, tmpdir):
 
 def get_hdf5_module(resource):
     """Return the correct h5py module based on the input resource."""
-    if resource.startswith(remote_h5):
-        return h5pyd
-    else:
-        return h5py
+    if isinstance(resource, str):
+        if resource.startswith(remote_h5):
```
shoyer: you can consolidate these into one condition with an `and`

h5netcdf/tests/test_h5netcdf.py
```diff
@@ -277,7 +279,10 @@ def read_legacy_netcdf(tmp_netcdf, read_module, write_module):
 
 
 def read_h5netcdf(tmp_netcdf, write_module):
-    remote_file = True if tmp_netcdf.startswith(remote_h5) else False
+    remote_file = False
+    if isinstance(tmp_netcdf, str):
+        if tmp_netcdf.startswith(remote_h5):
+            remote_file = True
```
shoyer: should be:
`remote_file = isinstance(tmp_netcdf, str) and tmp_netcdf.startswith(remote_h5)`

scottyhq: thanks @shoyer, all great recommendations!

h5netcdf/core.py
```diff
@@ -577,21 +579,30 @@ class File(Group):
...
+                        self._preexisting_file = True
+                    except IOError:
+                        self._preexisting_file = False
+                    self._h5file = h5pyd.File(path, mode, **kwargs)
+                else:
+                    self._preexisting_file = os.path.exists(path)
+                    self._h5file = h5py.File(path, mode, **kwargs)
+            else:  # file-like object
+                if h5py.__version__ < LooseVersion('2.9.0'):
                     raise ImportError(
```
shoyer: I'm not sure it really matters, but `ImportError` seems a little off here. Maybe `TypeError` would be more appropriate?

h5netcdf/tests/test_h5netcdf.py
```diff
@@ -277,7 +280,9 @@ def read_legacy_netcdf(tmp_netcdf, read_module, write_module):
 
 
 def read_h5netcdf(tmp_netcdf, write_module):
-    remote_file = True if tmp_netcdf.startswith(remote_h5) else False
+    remote_file = False
+    if isinstance(tmp_netcdf, str) and tmp_netcdf.startswith(remote_h5):
```
shoyer: I don't know why this code previously used a ternary conditional statement given that `True if x else False` is equivalent to `x`. Here I would definitely just assign the value of `remote_file` directly.

scottyhq:
Thanks for the suggestions @shoyer, this is much improved now. More tests could be helpful given many 'file-like' objects available (e.g. https://github.com/martindurant/filesystem_spec), but this illustrates the functionality. 

I'm mainly interested in opening files over networks, and there is a test in the accompanying xarray pull request https://github.com/pydata/xarray/pull/2782 for that. But I was having trouble coming across small netCDF4 files on a public server... Did I miss one in the pydata or xarray github repos? There are many HDF5 files out there (https://hdfeos.org/zoo/), but many of these encounter issues or errors when loading into an xarray dataset (e.g. https://github.com/shoyer/h5netcdf/issues/43).

shoyer:
I think it's fine to leave out extensive tests here. Presumably there is pretty decent coverage for file-like objects in h5py, and we aren't changing much here.

We have a couple of netCDF3 files in https://github.com/pydata/xarray-data but not netCDF4 files currently. They would be easy to add, though.

shoyer:
I don't know why tests are failing on Travis, but this passed locally on my machine.

This is now out in the 0.7.0 release

jhamman:
Nicely done @scottyhq! 🎉

PR #58, opened 2019-05-30
Disable h5pyd tests

shoyer:
These are failing on Travis-CI with "OSError: [Errno 404] Not Found", e.g., see https://travis-ci.org/shoyer/h5netcdf/jobs/539292517


shoyer:
Of course, there was actually a failing test with old versions of Python, now fixed :)

PR #60, opened 2019-07-01
Compatibility Python 3.8

paugier:
```
/home/users/me/Dev/h5netcdf/h5netcdf/dimensions.py:1: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
    from collections import MutableMapping
```

I didn't try to keep the compatibility with Python 2.7 since we are in July 2019 (see https://python3statement.org/).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,8 +5,6 @@ notifications:
 
 matrix:
   include:
-  - env: PYTHON_VERSION=2.7
-  - env: PYTHON_VERSION=3.4
   - env: PYTHON_VERSION=3.5
   - env: PYTHON_VERSION=3.6
   - env: PYTHON_VERSION=3.7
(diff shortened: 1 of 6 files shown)


shoyer:
Thanks for the PR!

Dropping Python 2.7 is fine (we could drop Python 3.4, too), but please also remove it from our CI setup so tests can pass:
https://github.com/shoyer/h5netcdf/blob/master/.travis.yml

Also it would be great to add `python_requires` into our `setup.py` file:
https://github.com/shoyer/h5netcdf/blob/master/setup.py

(should also drop the "Classifiers" for `Programming Language :: Python :: 2'`)

paugier:
Thanks for your comments. It should be better now!

I read that having 
```
if sys.version_info[:2] < (3, 5):
    raise RuntimeError('Python version >= 3.5 required.')
```
is a good practice to fail early.

PR #64, opened 2019-12-04
ENH: invent `phony_dim` names for unlabeled dimensions

kmuehlbauer:
This PR adds naming for unlabeled dimensions (without dimension scales) in a netcdf style manner (`phony_dim_0` etc) instead of raising an error. 

This works nice in most cases (hdf5 files with no dimension scales at all). In mixing cases the naming might need improvement. ~~I did not find a way of having an overall count of the invented `phony_dim_N`, so with the proposed changes it the counting is per dataset/subgroup~~.

Xref: https://github.com/shoyer/h5netcdf/issues/43

### DIFF
--- a/h5netcdf/core.py
+++ b/h5netcdf/core.py
@@ -1,5 +1,6 @@
 # For details on how netCDF4 builds on HDF5:
 # http://www.unidata.ucar.edu/software/netcdf/docs/file_format_specifications.html#netcdf_4_spec
+from collections import defaultdict
 from collections.abc import Mapping
 import os.path
 import warnings
@@ -95,16 +96,25 @@ class BaseVariable(object):
             return (child_name,)
 
         dims = []
+        phony_dims = defaultdict(int)
         for axis, dim in enumerate(self._h5ds.dims):
-            # TODO: read dimension labels even if there is no associated
-            # scale? it's not netCDF4 spec, but it is unambiguous...
-            # Also: the netCDF lib can read HDF5 datasets with unlabeled
-            # dimensions.
-            if len(dim) == 0:
-                raise ValueError('variable %r has no dimension scale '
-                                 'associated with axis %s'
-                                 % (self.name, axis))
-            name = _name_from_dimension(dim)
+            # get current dimension
+            dimsize = self.shape[axis]
+            phony_dims[dimsize] += 1
+            if len(dim):
+                name = _name_from_dimension(dim)
+            else:
+                # if unlabeled dimensions are found
+                if self._root._phony_dim_count is None:
+                    raise ValueError('variable %r has no dimension scale '
+                                     'associated with axis %s. \n'
+                                     'Use phony_dims=%r for sorted naming or '
+                                     'phony_dims=%r for per access naming.'
+                                     % (self.name, axis, 'sort', 'access'))
+                else:
+                    # get dimension name
+                    name = self._parent._phony_dims[(dimsize,
+                                                     phony_dims[dimsize] - 1)]
             dims.append(name)
         return tuple(dims)
 
@@ -211,6 +221,13 @@ def _netcdf_dimension_but_not_variable(h5py_dataset):
     return NOT_A_VARIABLE in h5py_dataset.attrs.get('NAME', b'')
 
 
+def _unlabeled_dimension_mix(h5py_dataset):
+    dims = sum([len(j) for j in h5py_dataset.dims])
+    if dims:
+        assert dims == h5py_dataset.ndim
+    return dims
+
+
 class Group(Mapping):
 
     _variable_cls = Variable
@@ -233,6 +250,12 @@ class Group(Mapping):
         self._variables = _LazyObjectLookup(self, self._variable_cls)
         self._groups = _LazyObjectLookup(self, self._group_cls)
 
+        # # initialize phony dimension counter
+        if self._root._phony_dim_count is not None:
+            self._phony_dims = {}
+            phony_dims = defaultdict(int)
+            labeled_dims = defaultdict(int)
+
         for k, v in self._h5group.items():
             if isinstance(v, h5_group_types):
                 # add to the groups collection if this is a h5py(d) Group
@@ -253,6 +276,12 @@ class Group(Mapping):
                         current_size = v.size
 
                     self._dim_sizes[k] = size
+
+                    # keep track of found labeled dimensions
+                    if self._root._phony_dim_count is not None:
+                        labeled_dims[size] += 1
+                        self._phony_dims[(size, labeled_dims[size] - 1)] = k
+
                     # Figure out the current size of a dimension, which for
                     # unlimited dimensions requires looking at the actual
                     # variables.
@@ -260,6 +289,18 @@ class Group(Mapping):
                         self._determine_current_dimension_size(k, current_size)
 
                     self._dim_order[k] = dim_id
+                else:
+                    if self._root._phony_dim_count is not None:
+                        # check if malformed variable
+                        if not _unlabeled_dimension_mix(v):
+                            # if unscaled variable, get phony dimensions
+                            vdims = defaultdict(int)
+                            for i in v.shape:
+                                vdims[i] += 1
+                            for dimsize, cnt in vdims.items():
+                                phony_dims[dimsize] = max(phony_dims[dimsize],
+                                                          cnt)
+
                 if not _netcdf_dimension_but_not_variable(v):
                     if isinstance(v, h5py.Dataset):
                         var_name = k
@@ -267,7 +308,35 @@ class Group(Mapping):
                             var_name = k[len('_nc4_non_coord_'):]
                         self._variables.add(var_name)
 
+        # iterate over found phony dimensions and create them
+        if self._root._phony_dim_count is not None:
+            grp_phony_count = 0
+            for size, cnt in phony_dims.items():
+                # only create missing dimensions
+                for pcnt in range(labeled_dims[size], cnt):
+                    name = (grp_phony_count +
+                            self._root._phony_dim_count)
+                    grp_phony_count += 1
+                    # this is for per access naming
+                    if self._root._labeled_dim_count is None:
+                        name = "phony_dim_{}".format(name)
+                        self._create_dimension(name, size)
+                    self._phony_dims[(size, pcnt)] = name
+            # finally increase phony dim count at file level
+            self._root._phony_dim_count += grp_phony_count
+
         self._initialized = True
+
+    def _create_phony_dimensions(self):
+        # this is for 'sort' naming
+        for key, value in self._phony_dims.items():
+            try:
+                value += self._root._labeled_dim_count
+                name = "phony_dim_{}".format(value)
+                self._create_dimension(name, key[0])
+                self._phony_dims[key] = name
+            except TypeError:
+                pass
 
     def _determine_current_dimension_size(self, dim_name, max_size):
         """
@@ -582,7 +651,8 @@ class Group(Mapping):
 
 class File(Group):
 
-    def __init__(self, path, mode='a', invalid_netcdf=None, **kwargs):
+    def __init__(self, path, mode='a', invalid_netcdf=None, phony_dims=None,
+                 **kwargs):
         try:
             if isinstance(path, str):
                 if path.startswith(('http://', 'https://', 'hdf5://')):
@@ -622,6 +692,23 @@ class File(Group):
         # if we actually use invalid NetCDF features.
         self._write_ncproperties = invalid_netcdf is not True
 
+        # phony dimension handling
+        if phony_dims is not None:
+            if phony_dims == 'sort':
+                self._phony_dim_count = 0
+                self._labeled_dim_count = 0
+            elif phony_dims == 'access':
+                self._phony_dim_count = 0
+                self._labeled_dim_count = None
+            else:
+                raise ValueError('unknown value %r for phony_dims\n'
+                                 'Use phony_dims=%r for sorted naming, '
+                                 'phony_dims=%r for per access naming.'
+                                 % (phony_dims, 'sort', 'access'))
+        else:
+            self._phony_dim_count = None
+            self._labeled_dim_count = None
+
         # These maps keep track of dimensions in terms of size (might be
         # unlimited), current size (identical to size for limited dimensions),
         # their position, and look-up for HDF5 datasets corresponding to a
@@ -630,8 +717,26 @@ class File(Group):
         self._current_dim_sizes = ChainMap()
         self._dim_order = ChainMap()
         self._all_h5groups = ChainMap(self._h5group)
-
         super(File, self).__init__(self, self._h5path)
+        # initialize all groups to detect/create phony dimensions
+        # mimics netcdf-c style naming
+        if phony_dims == 'sort':
+            self._determine_phony_dimensions()
+
+    def _determine_phony_dimensions(self):
+        def get_labeled_dimension_count(grp):
+            count = len(grp._dim_sizes.maps[0])
+            for name in grp.groups:
+                count += get_labeled_dimension_count(grp[name])
+            return count
+
+        def create_phony_dimensions(grp):
+            grp._create_phony_dimensions()
+            for name in grp.groups:
+                create_phony_dimensions(grp[name])
+
+        self._labeled_dim_count = get_labeled_dimension_count(self)
+        create_phony_dimensions(self)
 
     def _check_valid_netcdf_dtype(self, dtype, stacklevel=3):
         dtype = np.dtype(dtype)
--- a/h5netcdf/core.py
+++ b/h5netcdf/core.py
@@ -105,7 +105,7 @@ class BaseVariable(object):
                 name = _name_from_dimension(dim)
             else:
                 # if unlabeled dimensions are found
-                if self._root._phony_dim_count is None:
+                if self._root._phony_dims_mode is None:
                     raise ValueError('variable %r has no dimension scale '
                                      'associated with axis %s. \n'
                                      'Use phony_dims=%r for sorted naming or '
@@ -254,7 +254,7 @@ class Group(Mapping):
         self._groups = _LazyObjectLookup(self, self._group_cls)
 
         # # initialize phony dimension counter
-        if self._root._phony_dim_count is not None:
+        if self._root._phony_dims_mode is not None:
             self._phony_dims = {}
             phony_dims = defaultdict(int)
             labeled_dims = defaultdict(int)
@@ -281,7 +281,7 @@ class Group(Mapping):
                     self._dim_sizes[k] = size
 
                     # keep track of found labeled dimensions
-                    if self._root._phony_dim_count is not None:
+                    if self._root._phony_dims_mode is not None:
                         labeled_dims[size] += 1
                         self._phony_dims[(size, labeled_dims[size] - 1)] = k
 
@@ -293,7 +293,7 @@ class Group(Mapping):
 
                     self._dim_order[k] = dim_id
                 else:
-                    if self._root._phony_dim_count is not None:
+                    if self._root._phony_dims_mode is not None:
                         # check if malformed variable
                         if not _unlabeled_dimension_mix(v):
                             # if unscaled variable, get phony dimensions
@@ -312,7 +312,7 @@ class Group(Mapping):
                         self._variables.add(var_name)
 
         # iterate over found phony dimensions and create them
-        if self._root._phony_dim_count is not None:
+        if self._root._phony_dims_mode is not None:
             grp_phony_count = 0
             for size, cnt in phony_dims.items():
                 # only create missing dimensions
@@ -320,8 +320,7 @@ class Group(Mapping):
                     name = (grp_phony_count +
                             self._root._phony_dim_count)
                     grp_phony_count += 1
-                    # this is for per access naming
-                    if self._root._labeled_dim_count is None:
+                    if self._root._phony_dims_mode == "access":
                         name = "phony_dim_{}".format(name)
                         self._create_dimension(name, size)
                     self._phony_dims[(size, pcnt)] = name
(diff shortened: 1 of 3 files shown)


kmuehlbauer:
Hmm, seems that the restapi tests are failing. Not sure how to tackle this, @shoyer. Your expertise is very much appreciated.

kmuehlbauer:
OK, seems that replacing `pytest.config` with `request.config` in the tests did the trick. But all remote tests are skipped. Not sure, if this is intended.

shoyer:
It's fine to skip the restapi tests.

shoyer:
This will need more unit tests, specifically to handle the case of multiple variables with possibly multiple unlabelled dimensions. The phony dimension names should be verified to be consistent with netcdf-C. You might need to store generated dimension names somewhere on group or file objects.

shoyer:
But thanks for working this, it will be a welcome addition!

kmuehlbauer:
@shoyer I totally agree that a complete compatibility with netcdf in terms of naming would be best. I'm having a bit of a hard time to find my way through the code. Should I try to use these chained objects to get store and retrieve the names?

shoyer:
The chained objects are useful for coordinating across a hierarchy of groups. Otherwise you can just use a normal dict on a Group object.

I don't know if netcdf coordinates phony dimension names across groups, or if they are assigned at the group level. I would suggest a little bit of experimentation to find out!

kmuehlbauer:
> I don't know if netcdf coordinates phony dimension names across groups, or if they are assigned at the group level. I would suggest a little bit of experimentation to find out!

netcdf-c creates a new phony dimension for every dimension found in any subgroup. So we need to handle this at the uppermost level (file-level). netcdf is crawling the whole file once and determining the needed phony dimensions.

Currently I'd like to create the phony dimensions here at the Variable level to keep laziness:
https://github.com/shoyer/h5netcdf/blob/f71e4a692bc11689a3ad7850e8ce4ac1cb536659/h5netcdf/core.py#L97-L110

But this will create the phony dimensions only if this group/variable is accessed. And this depends on the user. I'm currently thinking about the time overhead to traverse the whole file contents to iterate over all group/variables to create the dimensions at the right places with the right names.

I'm also a bit unsure how to correctly access the private parent methods/properties.

kmuehlbauer:
Behaviour in groups with multiple variables is now in line with latest netcdf style (see https://github.com/Unidata/netcdf-c/issues/1484 and https://github.com/Unidata/netcdf-c/pull/1525).

Unsolved problem is still the unknown order of accessing variables. Not sure how to proceed here with this.

kmuehlbauer:
@shoyer I think I'm ready with the creation of the phony dimensions. Before I start with adding tests, could you please share your thoughts on the lazy creation of the phony dimensions vs. creation at initialization?

kmuehlbauer:
Not sure, if this is elegant. The latest commit adds the iteration over the groups/variables (at file init-time) acquiring the `Variable.ndim` and leaving `._unlabeled_id` at the group-level.  The phony dimension is then first created with the saved id on access time. 

Seems, that this handling breaks some legacy things. Any hints appreciated.

kmuehlbauer:
With the last two commits, the determination routine for unscaled dimensions is only run the first time an unscaled dimension is detected. This moves the time-consuming part (adding 150ms on my test dataset, to the first access). This means only files which have at least one unscaled dimension are processed that way. All other files are unaffected.

Not sure, if the the two needed instance variables are OK (`.unscaled_dim_count` at File level, `.first_unscaled_dim_id` for the group level). With no unscaled dimensions they are both `None`.

Starting to add tests as suggested.

kmuehlbauer:
@shoyer I'm currently wrapping my head around keeping track of the created dimensions. Something is not correct, yet.

kmuehlbauer:
After letting this sink in for a while, it just is so plain simple. We need to sort the dict keys in Python < 3.7. Seems that this works now for a "simple" file with no dimensions defined at all. I'll add a test with mixed layout, too.

kmuehlbauer:
@shoyer This is (finally) ready for review.

shoyer:
Hi @kmuehlbauer, thanks for looking into this!

This does indeed seem to be pretty tricky when considering lazy loading of variables. One challenge is that netCDF4 seems to always include phony dimensions in `dimensions` on a Group. If we want to keep that consistent, then we will need to always inspect all variables in a Group to find any phony dimensions when `Group.dimensions` is accessed, not just when the first `Variable` with phony dimensions is loaded. I am not sure about the performance implications of this, but this is probably acceptable. At least we don't have to look at subgroups!

As for the logic inside your current implementation, I wonder if we might be able to simplify it by adapting a slightly different data model, by explicitly storing phony dimensions on `Group` objects as a mapping from size to dimension name, e.g., `{200: 'phony_dim_0', 300: 'phony_dim_1'}` where the names are determined based on the order in which the phony dimensions appear on variables.

kmuehlbauer:
> This does indeed seem to be pretty tricky when considering lazy loading of variables. One challenge is that netCDF4 seems to always include phony dimensions in `dimensions` on a Group. If we want to keep that consistent, then we will need to always inspect all variables in a Group to find any phony dimensions when `Group.dimensions` is accessed, not just when the first `Variable` with phony dimensions is loaded. I am not sure about the performance implications of this, but this is probably acceptable. At least we don't have to look at subgroups!

Thanks for the input. Indeed, that's worth exploring. I'll give this a try.

> As for the logic inside your current implementation, I wonder if we might be able to simplify it by adapting a slightly different data model, by explicitly storing phony dimensions on `Group` objects as a mapping from size to dimension name, e.g., `{200: 'phony_dim_0', 300: 'phony_dim_1'}` where the names are determined based on the order in which the phony dimensions appear on variables.

I'm not sure I get this. The problematic thing is if two dimensions in the same variable are of same size. If we think about a 2D array (time, measurement), where measurement is some 1D array (say of length 10). If the file contains:

- 5 timesteps -> (5, 10) -> `{5: 'phony_dim_0', 10: 'phony_dim_1'}`
- 10 timesteps -> (10, 10) -> `{10: 'phony_dim_0', 10: 'phony_dim_1'}` (doesn't work)

I've currently no other means to handle this, but maybe you have some more ideas?

kmuehlbauer:
@shoyer While digging into this, I found [this h5py issue](https://github.com/h5py/h5py/issues/744), which is directly related to these lines of code:

https://github.com/shoyer/h5netcdf/blob/f71e4a692bc11689a3ad7850e8ce4ac1cb536659/h5netcdf/core.py#L98-L106

The for-loop can should be written (without enumerate) as:

```python
 for axis, dim in self._h5ds.dims.items():
```
Seems like the `enumerate` is some workaround the mentioned issue. How should I move forward?

shoyer:
> I'm not sure I get this. The problematic thing is if two dimensions in the same variable are of same size. If we think about a 2D array (time, measurement), where measurement is some 1D array (say of length 10). If the file contains:
> 
> * 5 timesteps -> (5, 10) -> `{5: 'phony_dim_0', 10: 'phony_dim_1'}`
> * 10 timesteps -> (10, 10) -> `{10: 'phony_dim_0', 10: 'phony_dim_1'}` (doesn't work)
> 
> I've currently no other means to handle this, but maybe you have some more ideas?

netCDF-C seems to reuse a single phony dimension name if multiple dimensions have the same size:
```
<class 'netCDF4._netCDF4.Variable'>
float64 foo(phony_dim_0, phony_dim_0, phony_dim_1)
unlimited dimensions:
current shape = (2, 2, 3)
filling off
```

I agree that the logic seems questionable, but it is probably better to be consistent with netCDF4-Python than to invent our own convention.

shoyer:
> @shoyer While digging into this, I found [this h5py issue](https://github.com/h5py/h5py/issues/744), which is directly related to these lines of code:
> 
> https://github.com/shoyer/h5netcdf/blob/f71e4a692bc11689a3ad7850e8ce4ac1cb536659/h5netcdf/core.py#L98-L106
> 
> The for-loop can should be written (without enumerate) as:
> 
> ```python
>  for axis, dim in self._h5ds.dims.items():
> ```
> 
> Seems like the `enumerate` is some workaround the mentioned issue. How should I move forward?

You are probably right. I do not remember why this is code is how it is.

kmuehlbauer:
@shoyer Regarding reuse of phony dims this has changed with netcdf-c 4.7.3, see https://github.com/Unidata/netcdf-c/issues/1484

shoyer:
OK, we can certainly match the new netcdf-C behavior. In that case, I would suggest using a dictionary keyed by a tuple `(size, position)` where position is the number of times the dimension has appeared so far on a variable.

kmuehlbauer:
OK, I'll try as suggested.

kmuehlbauer:
@shoyer Thanks for your great insights. Now everything looks more readable. 

The determination and creation of the phony dimensions is now at group level, where the `_phony_dims` are saved as you suggested for use at variable lookup-time.

The penalty is the need to iterate all groups at `File` init-time. The phony dimension count is constantly updated while iterating and finally holds the overall phony dimension count.

As long as  https://github.com/h5py/h5py/issues/744 is not resolved we need to stick to the current `enumerate` workaround.

h5netcdf/core.py
```diff
@@ -223,6 +227,8 @@ def __init__(self, parent, name):
         self._parent = parent
         self._root = parent._root
         self._h5path = _join_h5paths(parent.name, name)
+        self._unscaled_dim_id = None
```
kmuehlbauer: Note to self: `self._unscaled_dim_id` need to be removed.

kmuehlbauer:
@shoyer FYI: Using `enumerate` with `dims` is correct. h5py is about to remove the `MappingHDF5` interface from `DimensionManager` class (https://github.com/h5py/h5py/pull/1464).

kmuehlbauer:
This isn't good, the performance is really bad, when iterating over all groups at file init and creating all possible  phony dims at the same time. I'll need to refactor this back into the variable lookup.

kmuehlbauer:
@shoyer Finally I'm really unsure what might be the best phony dimension handling. For the discovery of the phony dimensions I think this is more readable than the first approaches.

To be 100% compatible with netcdf-c (only naming) we would need to iterate all groups once which is a huge performance problem if you have many groups/subgroups with unscaled dimensions (in my use case > 300 per file). My best suggestion would be to create the dimensions on access. In case the user iterates in netcdf-c order over all groups/subgroups the same naming is achieved. If the user iterates in another fashion, well, the naming will be different.  My second best idea would be to intridoce yet another kwarg which switches this behaviour. WDYT?

(I've just seen that I need to sort the dict again for py36, will do the next day.)

kmuehlbauer:
This is now ready for another review-iteration, @shoyer. I hope all issues have been addressed properly.

kmuehlbauer:
Happy New Year! I'd be happy to work on ironing out any further issues her in this PR.

shoyer:
> My best suggestion would be to create the dimensions on access. In case the user iterates in netcdf-c order over all groups/subgroups the same naming is achieved. If the user iterates in another fashion, well, the naming will be different. My second best idea would be to intridoce yet another kwarg which switches this behaviour. WDYT?

I don't like behavior that depends on the order of accessing variables. That feels very surprising to me.

So I would lean towards the second solution here, introducing a switch to control this at the `File` level.

kmuehlbauer:
> I don't like behavior that depends on the order of accessing variables. That feels very surprising to me.
> 
> So I would lean towards the second solution here, introducing a switch to control this at the `File` level.

Taking your comment into account, I added the keyword `phony_dims`.

- `phony_dims=None` keeps the current behaviour to raise a ValueError if unlabeled dimensions are detected. I adapted the ErrorMessage pointing to the keyword.
- `phony_dims='sort'` creates the phony dimensions like the netcdf-c counterpart
- `phony_dims='access'` (or any other content except `None` or `sort`) creates the phony dimensions on the fly when accessing the respective groups.

I also added to the tests to catch the three different behaviours.

kmuehlbauer:
@shoyer The error in the Python3.5 build is due to `h5pyd` package using f-strings, so it's not correlated with the changes here.

kmuehlbauer:
@shoyer I rebased this on latest master now. Now we can see effects on Python3.5 with dictionary elements not in order of creation. I'll try to get this cleaned out.

shoyer:
We could also stop supporting Python 3.5 at this point -- it's very old now
and anyone who cares can continue to use old versions of this library.

kmuehlbauer:
> We could also stop supporting Python 3.5 at this point

If you are comfortable with this, I can remove the Python3.5 from the build matrix and add Python3.8 instead. Will have a PR ready shortly.

kmuehlbauer:
@shoyer see #67

kmuehlbauer:
OK, if abandoning Python3.5 is settled, then I think this is really ready now.

kmuehlbauer:
Hopefully the final rebase, @shoyer

kmuehlbauer:
@shoyer I know this has been a long road and I'm grateful for your guidance here. I think this PR is quite mature now and that your comments and suggestions have been handled accordingly. Is there anything I can do to get this to a merge-able state? I know that free cycles are sometimes hard to spare. Thanks again for your help!

Just a bit of background from my side. This PR and https://github.com/pydata/xarray/pull/3618 over at xarray are the very last steps of a very long road to make xarray able to read weather radar data in ODIMH5 format (hdf5 based format) with both netcdf4/h5netcdf engine. ODIMH5 is a standard used by weather services throughout the world (eg. in Europe, Australia, Canada). Hopefully I can further extend my knowledge on the insides of h5netcdf/xarray to help maintain these great libraries.

h5netcdf/core.py
```diff
@@ -260,12 +273,32 @@ def __init__(self, parent, name):
...
+            self._phony_dims = {}
+            for size in phony_dims:
+                for cnt in range(phony_dims[size]):
+                    name = 'phony_dim_{}'.format(
+                        len(self._phony_dims) +
+                        self._root._phony_dim_count)
+                    self._phony_dims[(size, cnt)] = name
+                    self._create_dimension(name, size)
+            # finally increase phony dim count
+            self._root._phony_dim_count += len(self._phony_dims)
```
shoyer: I notice that this keeps track of unique phony dimension name at the `File` rather than the `Group` level.

This seems reasonable, but just to verify, is this consistent with what netCDF4-Python does?

h5netcdf/core.py
```diff
@@ -95,16 +96,25 @@ def _lookup_dimensions(self):
...
+            if len(dim):
+                name = _name_from_dimension(dim)
+            else:
+                # if unlabeled dimensions are found
+                if self._root._phony_dim_count is None:
+                    raise ValueError('variable %r has no dimension scale '
+                                     'associated with axis %s. \n'
+                                     'Use phony_dims=%r for sorted naming or '
+                                     'phony_dims=%r for per access naming.'
+                                     % (self.name, axis, 'sort', 'access'))
```
shoyer: Do you have a use-case for "per access" naming?

If not, I would rather leave it out to keep things slightly simpler.

h5netcdf/core.py
```diff
@@ -95,16 +96,25 @@ def _lookup_dimensions(self):
             return (child_name,)
 
         dims = []
+        phony_dims = Counter()
```
shoyer: Nit (feel free to ignore): `Counter` is fine, but it is a little slow and does a lot of magic. I usually stick to the simpler `collections.defaultdict(int)`.

h5netcdf/core.py
```diff
@@ -622,12 +656,25 @@ def __init__(self, path, mode='a', invalid_netcdf=None, **kwargs):
...
         self._dim_sizes = ChainMap()
         self._current_dim_sizes = ChainMap()
         self._dim_order = ChainMap()
         self._all_h5groups = ChainMap(self._h5group)
-
         super(File, self).__init__(self, self._h5path)
+        # initialize all groups to detect/create phony dimensions
+        # mimics netcdf-c style naming
+        if phony_dims == 'sort':
+            self._determine_phony_dimensions()
```
shoyer: This should check to make use `phony_dims == 'access'` otherwise (or raise an error if not). Otherwise typos don't get checked.

h5netcdf/core.py
kmuehlbauer: AFAIK netcdf-c does all the hard work and netCDF4-Python just picks that up.  netcdf-c creates phony dims when iterating over all groups in a sorted manner. So yes, the current implementation mimicks the netcdf behaviour. 

Here we need to keep track of the count of the already created phony dimensions. This has to be on `File` level to get access from every subgroup.

I'll think about a test to compare netcdf4 and h5netcdf access.

h5netcdf/core.py
kmuehlbauer: Actually this is my use case. While working on this I found that iterating over all groups (with `phony_dims='sort'`) terribly affects performance. When using the `per access` approach this phony dimension detection is deferred to the time when data is accessed. This really improves in cases when working with hundreds or thousands of data files (eg. one day of a single radar of German Weather Service has > 50.000 files).

h5netcdf/core.py
kmuehlbauer: Everything which positively affects performance is better. I'll try to get your suggestion working.

h5netcdf/core.py
kmuehlbauer: Sure, will add the check.

h5netcdf/core.py
shoyer: I guess this would be a pretty strange dataset, but you could make a file where two different groups have variables with unknown dimensions that appear in the different orders.

h5netcdf/core.py
shoyer: OK, cool. I just wanted to be sure we were adding this feature for a clear reason.

kmuehlbauer:
@shoyer Everything works fine for the unlabeled only case. I'm currently struggling with the mixed case, when we have a mix of labeled and unlabeled dimensions. I've no idea, if such files even exist out there. The problem is that `netcdf-c` treats labeled dimensions first (assigning some id) for all (sub)groups and then all unlabeled dimensions for all subgroups counting up from that id. This currently fails with my groupwise creation of the phony dimensions. I'll dig a bit more to find a way to get around.

kmuehlbauer:
@shoyer This took some time to figure out.

Here is the final layout:

We check for the `phony_dim` keyword to be either `sort` or `access` or `None` (which is default). 
- If `None` we raise an error at variable access if unlabeled dimensions are detected and give some explanation on the use of the keyword.
- If `sort` we iterate once over the groups and keep track of the labeled and unlabeled dimensions at group level and for the overall counts at file level. Then we call new group function `_create_phony_dimensions` which create the final name-strings from the tracked values and calls `create_dimension`. This behaviour is netcdf-c compatible, but relatively slow.
- If `access` we create phony_dimensions at group access time, where the naming is not netcdf-c compatible but more performant

We also raise an error if malformed variable dims (mixed labeled and unlabeled dimensions) are detected, because this also breaks in netcdf-c.

I had to rebase on latest master due to the conflicts. Please let me know, if this can be further improved.

h5netcdf/core.py
```diff
@@ -253,22 +279,68 @@ def __init__(self, parent, name):
...
+    def _create_phony_dimensions(self):
+        # this is for 'sort' naming
+        for key, value in self._phony_dims.items():
+            try:
+                value += self._root._labeled_dim_count
+                name = "phony_dim_{}".format(value)
+                self._create_dimension(name, key[0])
+                self._phony_dims[key] = name
+            except TypeError:
+                pass
```
shoyer: This is a rather broad exception to be catching over multiple lines of code that could hide programming errors. When can this arise?

h5netcdf/core.py
```diff
@@ -622,6 +695,23 @@ def __init__(self, path, mode='a', invalid_netcdf=None, **kwargs):
...
         self._write_ncproperties = invalid_netcdf is not True
 
+        # phony dimension handling
+        if phony_dims is not None:
+            if phony_dims == 'sort':
+                self._phony_dim_count = 0
+                self._labeled_dim_count = 0
+            elif phony_dims == 'access':
+                self._phony_dim_count = 0
+                self._labeled_dim_count = None
```
shoyer: Encoding the phony dimension mode in the type of these arguments feels a little too clever/indirect to me. It is not obvious that `labeled_dim_count is None` means "per access" mode.

Can you store the phony dimension mode explicitly in an attribute, e.g., `self._phony_dims_mode = phony_dims`? I think that would make the other code a little easier to read.

h5netcdf/core.py
```diff
@@ -211,6 +221,16 @@ def _netcdf_dimension_but_not_variable(h5py_dataset):
...
 
+def _unlabeled_dimension_mix(h5py_dataset):
+    dims = sum([len(j) for j in h5py_dataset.dims])
+    if dims:
+        if dims != h5py_dataset.ndim:
+            name = h5py_dataset.name.split('/')[-1]
+            raise ValueError('malformed variable %r has mixing of labeled and '
+                             'unlabeled dimensions.'.format(name))
+    return dims
+
```
shoyer: I appreciate your attention to detail checking for these edge cases!

shoyer:
> @shoyer This took some time to figure out.
> 
> Here is the final layout:

This looks great to me! I have a few minor implementation notes but I think this is basically ready.

Now that we've figured out the interface, could you please also add a brief note to the README file? We might also note this in the changelog -- this will be the 0.8 release.

h5netcdf/core.py
kmuehlbauer: I could check the `value` with isinstance(int, value) which would have the same effect. Would that be OK?

h5netcdf/core.py
kmuehlbauer: Yes, this makes absolutely sense. I'll add this.

h5netcdf/core.py
shoyer: Yes, that would be much better!

kmuehlbauer:
> Now that we've figured out the interface, could you please also add a brief note to the README file? We might also note this in the changelog -- this will be the 0.8 release.

I've added the requested changes, indeed it's more readable and looks really nice now.

I will add the section to the readme and can also add a section to the changelog.

kmuehlbauer:
@shoyer I've added a README section and a changelog item for 0.8.0. I left the date open. If you have a date already in mind I can add it, or you can do this before release.

kmuehlbauer:
@shoyer Anything more I can do for having this merged?

kmuehlbauer:
@shoyer Since xarray 0.15.0 is out it would be really great if this also gets merged and released as h5netcdf 0.8.0. I've a release pending for one of my packages and this is the last piece here. If I can anything do to speed things up, please let me know.

shoyer:
thanks @kmuehlbauer, especially for your patience here!

I'll try to get a 0.8.0 release out shortly...

shoyer:
OK, I think I issued the release successfully on pypi. Let me know if that works for you.

kmuehlbauer:
@shoyer Huge thanks for releasing this. I'll test first thing tomorrow and report back.

kmuehlbauer:
@shoyer Works nicely at my place. Again, thanks for working through this with me!

PR #67, opened 2020-01-10
MNT: remove Python3.5 / add Python3.8 in travis.yml

kmuehlbauer:
- as suggested in #64 remove Python3.5 from build matrix
- also add Python3.8 to build matrix

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,9 @@ notifications:
 
 matrix:
   include:
-  - env: PYTHON_VERSION=3.5 H5PYD_VERSION=v0.5.0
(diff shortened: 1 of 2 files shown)


kmuehlbauer:
@ping shoyer Seems to work nicely with Python3.8.

shoyer:
Great, thank you! Would you mind also adding `python_requires` to setup.py as well? That stops people from installing on older versions of Python.

kmuehlbauer:
> Would you mind also adding `python_requires` to setup.py as well?

[x] Done

shoyer:
Thanks!

PR #82, opened 2020-11-10
FIX: adapt to string reading in h5py >= 3.0

### LINKED ISSUE #81: string handling changed in h5py >= 3.0
In `h5py` storing of strings has changed (see https://docs.h5py.org/en/latest/whatsnew/3.0.html and https://docs.h5py.org/en/latest/strings.html).

This currently breaks the testsuite and also downstream packages using `h5py>=3.0` (https://github.com/pydata/xarray/issues/4570).

Below is the `ncdump` of the three write methods used in the testsuit (using `h5py=3.1`), where the output of `netcdf4` differs from the output of the two `h5netcdf` flavours. 

<br/>
write module netcdf4:
<details>

```
write_module <module 'netCDF4' from '/home/kai/miniconda3/envs/h5py3/lib/python3.9/site-packages/netCDF4/__init__.py'>
netcdf testfile {
dimensions:
	x = 4 ;
	y = 5 ;
	z = 6 ;
	empty = UNLIMITED ; // (0 currently)
	string3 = 3 ;
	mismatched_dim = 1 ;
variables:
	double foo(x, y) ;
		foo:units = "meters" ;
	int64 y(y) ;
		y:_FillValue = -1LL ;
	char z(z, string3) ;
		z:_FillValue = "X" ;
	float scalar ;
	int64 intscalar ;
	int64 mismatched_dim ;
	string var_len_str(x) ;

// global attributes:
		:global = 42LL ;
		:other_attr = "yes" ;
data:

 foo =
  1, 1, 1, 1, 1,
  1, 1, 1, 1, 1,
  1, 1, 1, 1, 1,
  1, 1, 1, 1, 1 ;

 y = 0, 1, 2, 3, _ ;

 z =
  "a",
  "b",
  "c",
  "foo",
  "bar",
  "baz" ;

 scalar = 2 ;

 intscalar = 2 ;

 mismatched_dim = _ ;

 var_len_str = "foo", _, _, _ ;

group: subgroup {
  dimensions:
  	y = 10 ;
  variables:
  	int subvar(x) ;
  	double y_var(y) ;
  data:

   subvar = 0, 1, 2, 3 ;

   y_var = _, _, _, _, _, _, _, _, _, _ ;
  } // group subgroup
}
```
</details>

<br/>
write Module h5netcdf.legacy API:
<details>

```
write_module <module 'h5netcdf.legacyapi' from '/home/data/kai/projects/h5netcdf/h5netcdf/legacyapi.py'>
netcdf testfile {
dimensions:
	x = 4 ;
	y = 5 ;
	z = 6 ;
	empty = UNLIMITED ; // (0 currently)
	string3 = 3 ;
	mismatched_dim = 1 ;
variables:
	int64 mismatched_dim ;
	double foo(x, y) ;
		string foo:units = "meters" ;
	int64 intscalar ;
	float scalar ;
	string var_len_str(x) ;
	int64 y(y) ;
		y:_FillValue = -1LL ;
	char z(z, string3) ;
		z:_FillValue = "X" ;

// global attributes:
		:global = 42LL ;
		string :other_attr = "yes" ;
data:

 mismatched_dim = 0 ;

 foo =
  1, 1, 1, 1, 1,
  1, 1, 1, 1, 1,
  1, 1, 1, 1, 1,
  1, 1, 1, 1, 1 ;

 intscalar = 2 ;

 scalar = 2 ;

 var_len_str = "foo", NIL, NIL, NIL ;

 y = 0, 1, 2, 3, _ ;

 z =
  "a",
  "b",
  "c",
  "foo",
  "bar",
  "baz" ;

group: subgroup {
  dimensions:
  	y = 10 ;
  variables:
  	int subvar(x) ;
  	double y_var(y) ;
  data:

   subvar = 0, 1, 2, 3 ;

   y_var = 0, 0, 0, 0, 0, 0, 0, 0, 0, 0 ;
  } // group subgroup
}
```
</details>

<br/>
write Module h5netcdf API:
<details>

```
write_module: <module 'h5netcdf' from '/home/data/kai/projects/h5netcdf/h5netcdf/__init__.py'>
netcdf testfile {
dimensions:
	x = 4 ;
	y = 5 ;
	z = 6 ;
	empty = UNLIMITED ; // (0 currently)
	string3 = 3 ;
	mismatched_dim = 1 ;
variables:
	int64 mismatched_dim ;
	double foo(x, y) ;
		string foo:units = "meters" ;
	int64 intscalar ;
	float scalar ;
	string var_len_str(x) ;
	int64 y(y) ;
		y:_FillValue = -1LL ;
	char z(z, string3) ;
		z:_FillValue = "X" ;

// global attributes:
		:global = 42LL ;
		string :other_attr = "yes" ;
data:

 mismatched_dim = 0 ;

 foo =
  1, 1, 1, 1, 1,
  1, 1, 1, 1, 1,
  1, 1, 1, 1, 1,
  1, 1, 1, 1, 1 ;

 intscalar = 2 ;

 scalar = 2 ;

 var_len_str = "foo", NIL, NIL, NIL ;

 y = 0, 1, 2, 3, _ ;

 z =
  "a",
  "b",
  "c",
  "foo",
  "bar",
  "baz" ;

group: subgroup {
  dimensions:
  	y = 10 ;
  variables:
  	int subvar(x) ;
  	double y_var(y) ;
  data:

   subvar = 0, 1, 2, 3 ;

   y_var = 0, 0, 0, 0, 0, 0, 0, 0, 0, 0 ;
  } // group subgroup
}
```
</details>

A quick check using `h5py=2.10.0` leads to the same results. That means:

- There is a problem with writing which lead to different files. This does not break the testsuite using `h5py<3.0`.
- There is a problem with reading, which breaks the testsuite using `h5py>=3.0`.

One more note: Another indicator that only reading is the problem, is that the testsuite doesn't break for the tests where `netcdf4` is used for reading back files.

kmuehlbauer:
fixes #81 

How does it work?

- Keeps full backwards compatibility for `h5py` == 2.*

for `h5py` >= 3:

- If `legacy` API -> keeps conformance to netCDF4 (decoding strings by default, raise `TypeError` if `decode_strings` is explicitly set
- If `new` API -> keeps backwards compatibility for the moment (decoding strings by default, emit `FutureWarning` if `decode_strings` is not explicitly set, decoding can be deactivated by setting `decode_strings=False` )
- In future versions conformance with `h5py` >= 3 will be ensured by not decoding strings by default. String decoding can be requested by setting keyword argument `decode_strings=True`.

### DIFF
--- a/h5netcdf/core.py
+++ b/h5netcdf/core.py
@@ -151,14 +151,8 @@ class BaseVariable(object):
 
     def __getitem__(self, key):
         if getattr(self._root, "decode_strings", False):
-            try:
-                if h5py.check_string_dtype(self._h5ds[key].dtype) is not None:
-                    return self._h5ds.asstr()[key]
-            # need to catch object arrays of bytes (which are returned from
-            # HDF5 variable-length strings)
-            except AttributeError as err:
-                if err.args[0] == "'bytes' object has no attribute 'dtype'":
-                    return self._h5ds.asstr()[key]
+            if h5py.check_string_dtype(self._h5ds.dtype) is not None:
+                return self._h5ds.asstr()[key]
         return self._h5ds[key]
 
     def __setitem__(self, key, value):
--- a/h5netcdf/core.py
+++ b/h5netcdf/core.py
@@ -151,9 +151,14 @@ class BaseVariable(object):
 
     def __getitem__(self, key):
         if getattr(self._root, "decode_strings", False):
-            # need to wrap with np.array to fetch .dtype in any case
-            if h5py.check_string_dtype(np.array(self._h5ds[key]).dtype) is not None:
-                return self._h5ds.asstr()[key]
+            try:
+                if h5py.check_string_dtype(self._h5ds[key].dtype) is not None:
+                    return self._h5ds.asstr()[key]
+            # need to catch object arrays of bytes (which are returned from
+            # HDF5 variable-length strings)
+            except AttributeError as err:
+                if err.args[0] == "'bytes' object has no attribute 'dtype'":
+                    return self._h5ds.asstr()[key]
         return self._h5ds[key]
 
     def __setitem__(self, key, value):
--- a/h5netcdf/core.py
+++ b/h5netcdf/core.py
@@ -752,7 +752,7 @@ class File(Group):
             if "legacy" in self._cls_name:
                 if self.decode_strings is not None:
                     msg = (
-                        "'decode_strings' keyword argument is not allowed in h5netcdf"
+                        "'decode_strings' keyword argument is not allowed in h5netcdf "
                         "legacy API."
                     )
                     raise TypeError(msg)
--- a/h5netcdf/core.py
+++ b/h5netcdf/core.py
@@ -150,15 +150,10 @@ class BaseVariable(object):
         return self._h5ds.__array__(*args, **kwargs)
 
     def __getitem__(self, key):
-        if h5py.__version__ >= LooseVersion("3.0.0") and self._root.decode_strings:
-            if self.dtype == str:
+        if getattr(self._root, "decode_strings", False):
+            # need to wrap with np.array to fetch .dtype in any case
+            if h5py.check_string_dtype(np.array(self._h5ds[key]).dtype) is not None:
                 return self._h5ds.asstr()[key]
-            if self.dtype == object:
-                if isinstance(self._h5ds[key], bytes) or (
-                    isinstance(self._h5ds[key], np.ndarray)
-                    and type(self._h5ds[key].flat[0]) == bytes
-                ):
-                    return self._h5ds.asstr()[key]
         return self._h5ds[key]
 
     def __setitem__(self, key, value):
@@ -693,7 +688,7 @@ class Group(Mapping):
 class File(Group):
     def __init__(self, path, mode="a", invalid_netcdf=None, phony_dims=None, **kwargs):
         if h5py.__version__ >= LooseVersion("3.0.0"):
-            decode_strings = kwargs.pop("decode_strings", None)
+            self.decode_strings = kwargs.pop("decode_strings", None)
         try:
             if isinstance(path, str):
                 if path.startswith(("http://", "https://", "hdf5://")):
@@ -750,9 +745,14 @@ class File(Group):
         # string decoding
         if h5py.__version__ >= LooseVersion("3.0.0"):
             if "legacy" in self._cls_name:
+                if self.decode_strings is not None:
+                    msg = (
+                        "'decode_strings' keyword argument is not allowed in h5netcdf"
+                        "legacy API."
+                    )
+                    raise TypeError(msg)
                 self.decode_strings = True
             else:
-                self.decode_strings = decode_strings
                 if self.decode_strings is None:
                     msg = (
                         "String decoding changed with h5py >= 3.0. "
--- a/h5netcdf/core.py
+++ b/h5netcdf/core.py
@@ -150,6 +150,15 @@ class BaseVariable(object):
         return self._h5ds.__array__(*args, **kwargs)
 
     def __getitem__(self, key):
+        if h5py.__version__ >= LooseVersion("3.0.0") and self._root.decode_strings:
+            if self.dtype == str:
+                return self._h5ds.asstr()[key]
+            if self.dtype == object:
+                if isinstance(self._h5ds[key], bytes) or (
+                    isinstance(self._h5ds[key], np.ndarray)
+                    and type(self._h5ds[key].flat[0]) == bytes
+                ):
+                    return self._h5ds.asstr()[key]
         return self._h5ds[key]
 
     def __setitem__(self, key, value):
@@ -683,6 +692,8 @@ class Group(Mapping):
 
 class File(Group):
     def __init__(self, path, mode="a", invalid_netcdf=None, phony_dims=None, **kwargs):
+        if h5py.__version__ >= LooseVersion("3.0.0"):
+            decode_strings = kwargs.pop("decode_strings", None)
         try:
             if isinstance(path, str):
                 if path.startswith(("http://", "https://", "hdf5://")):
@@ -735,6 +746,23 @@ class File(Group):
                     "phony_dims=%r for per access naming."
                     % (phony_dims, "sort", "access")
                 )
+
+        # string decoding
+        if h5py.__version__ >= LooseVersion("3.0.0"):
+            if "legacy" in self._cls_name:
+                self.decode_strings = True
+            else:
+                self.decode_strings = decode_strings
+                if self.decode_strings is None:
+                    msg = (
+                        "String decoding changed with h5py >= 3.0. "
+                        "Currently backwards compatibility with h5py < 3.0 is kept by "
+                        "decoding strings per default. This will change in future "
+                        "versions for consistency with h5py >= 3.0. "
+                        "Setting 'decode_strings=True' forces string decoding."
+                    )
+                    warnings.warn(msg, FutureWarning, stacklevel=0)
+                    self.decode_strings = True
 
         # These maps keep track of dimensions in terms of size (might be
         # unlimited), current size (identical to size for limited dimensions),
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -39,7 +39,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.6", "3.7", "3.8", "3.9"]
+        python-version: ["3.7", "3.8", "3.9"]
     steps:
     - uses: actions/checkout@v2
     - uses: conda-incubator/setup-miniconda@v2
@@ -70,7 +70,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.6", "3.7", "3.8", "3.9"]
+        python-version: ["3.7", "3.8", "3.9"]
     steps:
     - uses: actions/checkout@v2
     - uses: conda-incubator/setup-miniconda@v2
--- a/h5netcdf/tests/test_h5netcdf.py
+++ b/h5netcdf/tests/test_h5netcdf.py
@@ -209,9 +209,8 @@ def write_h5netcdf(tmp_netcdf):
     ds.close()
 
 
-def read_legacy_netcdf(tmp_netcdf, read_module, write_module, decode_strings):
-
-    ds = read_module.Dataset(tmp_netcdf, "r", **decode_strings)
+def read_legacy_netcdf(tmp_netcdf, read_module, write_module):
+    ds = read_module.Dataset(tmp_netcdf, "r")
     assert ds.ncattrs() == ["global", "other_attr"]
     assert ds.getncattr("global") == 42
     if write_module is not netCDF4:
@@ -262,16 +261,19 @@ def read_legacy_netcdf(tmp_netcdf, read_module, write_module, decode_strings):
     # Check the behavior if h5py. Cannot expect h5netcdf to overcome these
     # errors:
     if is_h5py_char_working(tmp_netcdf, "z"):
-        ds = read_module.Dataset(tmp_netcdf, "r", **decode_strings)
+        ds = read_module.Dataset(tmp_netcdf, "r")
         v = ds.variables["z"]
-        assert array_equal(v, _char_array)
+        if getattr(ds, "decode_strings", False):
+            assert array_equal(v, np.char.decode(_char_array))
+        else:
+            assert array_equal(v, _char_array)
         assert v.dtype == "S1"
         assert v.ndim == 2
         assert v.dimensions == ("z", "string3")
         assert v.ncattrs() == ["_FillValue"]
         assert v.getncattr("_FillValue") == b"X"
     else:
-        ds = read_module.Dataset(tmp_netcdf, "r", **decode_strings)
+        ds = read_module.Dataset(tmp_netcdf, "r")
 
     v = ds.variables["scalar"]
     assert array_equal(v, np.array(2.0))
@@ -289,7 +291,7 @@ def read_legacy_netcdf(tmp_netcdf, read_module, write_module, decode_strings):
 
     v = ds.variables["var_len_str"]
     assert v.dtype == str
-    assert v[0] == u"foo"
+    assert v[0] == "foo"
 
     v = ds.groups["subgroup"].variables["subvar"]
     assert ds.groups["subgroup"].parent is ds
@@ -357,6 +359,10 @@ def read_h5netcdf(tmp_netcdf, write_module, decode_strings):
     if is_h5py_char_working(tmp_netcdf, "z"):
         ds = h5netcdf.File(tmp_netcdf, "r", **decode_strings)
         v = ds["z"]
+        if getattr(ds, "decode_strings", False):
+            assert array_equal(v, np.char.decode(_char_array))
+        else:
+            assert array_equal(v, _char_array)
         assert v.dtype == "S1"
         assert v.ndim == 2
         assert v.dimensions == ("z", "string3")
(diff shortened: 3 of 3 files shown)


kmuehlbauer:
@shoyer Seems that Travis-CI isn't running. Locally it works nicely and downstream xarray-testsuite also works with the proposed changes.

shaunc:
Will be happy to see this released! :)

kmuehlbauer:
@shaunc It might be necessary, to check if the object is really a string . 

Another problem is, unfortunately, that the Travis-CI hooks via travis-ci.org which got out of service recently and the repo isn't moved to travis-ci.com yet. 

We might need to wait until @shoyer can take a look at this.

shaunc:
@kmuehlbauer Well -- thanks for your work! I'm an Xarray user. Currently I have a wrapper function around my `open_data{set/array}` which reconverts strings -- obviously not ideal!

kmuehlbauer:
@shaunc Yeah, xarray-user (and part-time contributor) here too.

I would volunteer to help in maintaining `h5netcdf`.

shoyer:
@kmuehlbauer thanks for volunteering! I just sent you an invite to collaborate on this repo. We might also consider moving it over to a community organization, e.g., perhaps xarray-contrib?

kmuehlbauer:
@shoyer Thanks! I've just accepted the invitation. First thing would be to get CI back. To get Travis back running it would need to move the repo from travis-ci.org to travis-ci.com.

kmuehlbauer:
OK, seems the test [did actually run](https://travis-ci.org/github/shoyer/h5netcdf/builds/742667254), but are not linked here. I'll look into this the next days.

shoyer:
If you're up for it, I would suggest trying to migrate CI from Travis to GitHub Actions. It seems like Travis is not interested in supporting open source any longer, so we will probably have to do this eventually. The setup is pretty simple, all in one file: https://github.com/h5netcdf/h5netcdf/blob/master/.travis.yml

kmuehlbauer:
@shoyer I'll give GitHub Actions a go in my fork first. We'll see how this works out.

kmuehlbauer:
@shyoyer How should we move forward with this? Do you think the proposed fix will work in all cases, or should there be a more extensive test for string/type?

kmuehlbauer:
@shoyer @drew-parsons

I've implemented the solution discussed in #81. I'm not sure, if I'm overly strict with the conditionals. Any suggestions for improving are very much appreciated.

kmuehlbauer:
gentle nudge @shoyer

drew-parsons:
I've tested the patch on Debian.  It's working fine, test logs reported at
https://ci.debian.net/packages/p/python-h5netcdf/unstable/amd64/
https://ci.debian.net/packages/p/python-h5netcdf/testing/amd64/

kmuehlbauer:
@shoyer Is there anything left here to do, to finalize string handling for `h5py>=3.0`. I've tried to implement based on discussion in #81. We would also need to cut a release (0.9.0 ?) to make this available downstream (eg. `xarray`).

h5netcdf/core.py
```diff
@@ -150,6 +150,15 @@ def __array__(self, *args, **kwargs):
         return self._h5ds.__array__(*args, **kwargs)
 
     def __getitem__(self, key):
+        if h5py.__version__ >= LooseVersion("3.0.0") and self._root.decode_strings:
```
shoyer, line 153: This code can probably just check `decode_strings`?
```suggestion
        if self._root.decode_strings:
```

h5netcdf/core.py
```diff
@@ -736,6 +747,23 @@ def __init__(self, path, mode="a", invalid_netcdf=None, phony_dims=None, **kwarg
                     % (phony_dims, "sort", "access")
                 )
 
+        # string decoding
+        if h5py.__version__ >= LooseVersion("3.0.0"):
+            if "legacy" in self._cls_name:
+                self.decode_strings = True
```
shoyer, line 753: In this case, we should raise an `TypeError` if `decode_strings` was explicitly passed.

h5netcdf/core.py
```diff
@@ -150,6 +150,15 @@ def __array__(self, *args, **kwargs):
...
     def __getitem__(self, key):
+        if h5py.__version__ >= LooseVersion("3.0.0") and self._root.decode_strings:
+            if self.dtype == str:
+                return self._h5ds.asstr()[key]
+            if self.dtype == object:
+                if isinstance(self._h5ds[key], bytes) or (
+                    isinstance(self._h5ds[key], np.ndarray)
+                    and type(self._h5ds[key].flat[0]) == bytes
+                ):
+                    return self._h5ds.asstr()[key]
```
shoyer, line 161: I think this would work more reliable, by using h5py's API to check for string dtypes:
```suggestion
            if h5py.check_string_dtype(self._h5ds[key].dtype) is not None:
                return self._h5ds.asstr()[key]
```

shoyer:
thanks for your patience here on the review! please take a look at my comments

h5netcdf/core.py
kmuehlbauer: I've applied your suggestion using `getattr` since `decode_strings` isn't available in all cases.

h5netcdf/core.py
kmuehlbauer: Calling legacy API with `decode_strings` raises TypeError now.

h5netcdf/core.py
kmuehlbauer: This took a while. We need to wrap `self._h5ds[key]` with `np.array()` to evaluate `.dtype` in any case.

kmuehlbauer:
@shoyer I've update the PR per your suggestions. 

I've also simplified the tests. Now only the new API tests are run with `decode_strings = [True|False]`.

h5netcdf/core.py
```diff
@@ -750,9 +745,14 @@ def __init__(self, path, mode="a", invalid_netcdf=None, phony_dims=None, **kwarg
         # string decoding
         if h5py.__version__ >= LooseVersion("3.0.0"):
             if "legacy" in self._cls_name:
+                if self.decode_strings is not None:
+                    msg = (
+                        "'decode_strings' keyword argument is not allowed in h5netcdf"
```
shoyer, line 750:
```suggestion
                        "'decode_strings' keyword argument is not allowed in h5netcdf "
```

h5netcdf/core.py
```diff
@@ -150,15 +150,10 @@ def __array__(self, *args, **kwargs):
         return self._h5ds.__array__(*args, **kwargs)
 
     def __getitem__(self, key):
-        if h5py.__version__ >= LooseVersion("3.0.0") and self._root.decode_strings:
-            if self.dtype == str:
+        if getattr(self._root, "decode_strings", False):
+            # need to wrap with np.array to fetch .dtype in any case
+            if h5py.check_string_dtype(np.array(self._h5ds[key]).dtype) is not None:
```
shoyer, line 155: This is potentially problematic, because converting into a numpy array loads the entire HDF5 array into memory.

Can you give an example of why this fails?

kmuehlbauer: Ah, didn't think about that. The problem is that 'bytes' do not have dtype and then it breaks. That was the reason for the special checks for 'bytes'.

kmuehlbauer: So if `self._h5ds[key]` is of type `bytes`, then `self. _h5ds[key].dtype` raises `AttributeError`.

shoyer: > So if `self._h5ds[key]` is of type `bytes`

When can this happen? What does the underlying HDF5 file look like? As far as I can tell, that shouldn't happen...

kmuehlbauer: The problem is with the variable length string. 

Simple example to reproduce:

```python
    ds = netCDF4.Dataset(tmp_local_netcdf, "w")
    ds.createDimension("x", 4)
    v = ds.createVariable("var_len_str", str, ("x"))
    v[0] = u"foo"
    ds.close()

    ds = h5netcdf.legacyapi.Dataset(tmp_local_netcdf, "r")
    v = ds.variables["var_len_str"]
    assert v.dtype == str
    assert v[0] == u"foo"
```

Error traceback:
```
<h5netcdf.legacyapi.Variable '/var_len_str': dimensions ('x',), shape (4,), dtype <class 'str'>>
Attributes:
<HDF5 dataset "var_len_str": shape (4,), type "|O">
b'foo'

h5netcdf/tests/test_h5netcdf.py:120 (test_bytes)
tmp_local_netcdf = '/tmp/pytest-of-kai/pytest-8/test_bytes0/testfile.nc'

    def test_bytes(tmp_local_netcdf):
        ds = netCDF4.Dataset(tmp_local_netcdf, "w")
        ds.createDimension("x", 4)
        v = ds.createVariable("var_len_str", str, ("x"))
        v[0] = u"foo"
        ds.close()
    
        ds = h5netcdf.legacyapi.Dataset(tmp_local_netcdf, "r")
        v = ds.variables["var_len_str"]
        assert v.dtype == str
>       assert v[0] == u"foo"

h5netcdf/tests/test_h5netcdf.py:131: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <h5netcdf.legacyapi.Variable '/var_len_str': dimensions ('x',), shape (4,), dtype <class 'str'>>
Attributes:
key = 0

    def __getitem__(self, key):
        if getattr(self._root, "decode_strings", False):
            # need to wrap with np.array to fetch .dtype in any case
            try:
>               if h5py.check_string_dtype(self._h5ds[key].dtype) is not None:
E               AttributeError: 'bytes' object has no attribute 'dtype'

h5netcdf/core.py:156: AttributeError

```
h5dump of testfile.nc

```
(h5py3) kai@linux:~$ h5dump /tmp/pytest-of-kai/pytest-7/test_bytes0/testfile.nc
HDF5 "/tmp/pytest-of-kai/pytest-7/test_bytes0/testfile.nc" {
GROUP "/" {
   ATTRIBUTE "_NCProperties" {
      DATATYPE  H5T_STRING {
         STRSIZE 34;
         STRPAD H5T_STR_NULLTERM;
         CSET H5T_CSET_ASCII;
         CTYPE H5T_C_S1;
      }
      DATASPACE  SCALAR
      DATA {
      (0): "version=2,netcdf=4.7.4,hdf5=1.10.6"
      }
   }
   DATASET "var_len_str" {
      DATATYPE  H5T_STRING {
         STRSIZE H5T_VARIABLE;
         STRPAD H5T_STR_NULLTERM;
         CSET H5T_CSET_UTF8;
         CTYPE H5T_C_S1;
      }
      DATASPACE  SIMPLE { ( 4 ) / ( 4 ) }
      DATA {
      (0): "foo", "", "", ""
      }
      ATTRIBUTE "DIMENSION_LIST" {
         DATATYPE  H5T_VLEN { H5T_REFERENCE { H5T_STD_REF_OBJECT }}
         DATASPACE  SIMPLE { ( 1 ) / ( 1 ) }
         DATA {
         (0): (DATASET 331 /x )
         }
      }
   }
   DATASET "x" {
      DATATYPE  H5T_IEEE_F32BE
      DATASPACE  SIMPLE { ( 4 ) / ( 4 ) }
      DATA {
      (0): 0, 0, 0, 0
      }
      ATTRIBUTE "CLASS" {
         DATATYPE  H5T_STRING {
            STRSIZE 16;
            STRPAD H5T_STR_NULLTERM;
            CSET H5T_CSET_ASCII;
            CTYPE H5T_C_S1;
         }
         DATASPACE  SCALAR
         DATA {
         (0): "DIMENSION_SCALE"
         }
      }
      ATTRIBUTE "NAME" {
         DATATYPE  H5T_STRING {
            STRSIZE 64;
            STRPAD H5T_STR_NULLTERM;
            CSET H5T_CSET_ASCII;
            CTYPE H5T_C_S1;
         }
         DATASPACE  SCALAR
         DATA {
         (0): "This is a netCDF dimension but not a netCDF variable.         4"
         }
      }
      ATTRIBUTE "REFERENCE_LIST" {
         DATATYPE  H5T_COMPOUND {
            H5T_REFERENCE { H5T_STD_REF_OBJECT } "dataset";
            H5T_STD_I32LE "dimension";
         }
         DATASPACE  SIMPLE { ( 1 ) / ( 1 ) }
         DATA {
         (0): {
               DATASET 685 /var_len_str ,
               0
            }
         }
      }
      ATTRIBUTE "_Netcdf4Dimid" {
         DATATYPE  H5T_STD_I32LE
         DATASPACE  SCALAR
         DATA {
         (0): 0
         }
      }
   }
}
}
```

kmuehlbauer: Good chance there is some quirk in my logic.

kmuehlbauer: @shoyer The latest change catches the `AttributeError` and assumes that these `bytes` arrays should be treated with `.asstr()` too. This prevents the unconditional loading via the former `np.array()`-wrapper. I'm really not sure, there might be a better solution, though.

kmuehlbauer: @shoyer  Now, after thinking a bit more about that, we should test against the dtype of the dataset, which works quite nice.

kmuehlbauer:
@shoyer  After finally fixing the string check, everything looks good from my side. I've updated the OP with a short description of how this works.

h5netcdf/core.py
kmuehlbauer: Fixed by checking against `self._h5ds.dtype` instead of `self._h5ds[key].dtype`.

I've looked at these two lines of the traceback for quite some time until I got it:

```
        assert v.dtype == str
>       assert v[0] == u"foo"
```

shoyer: Ah, excellent! Yes, that's exactly what I was trying to suggest

shoyer (approved):

1ace:
Hi @kmuehlbauer!

I see you [removed Python 3.6 from the CI](https://github.com/h5netcdf/h5netcdf/pull/82/commits/5d5ca7f3cad3ec5238b3be0d08e343c991e7c849), does that mean that the fix in this PR does not work as intended on Python 3.6?

[`setup.py`](https://github.com/h5netcdf/h5netcdf/blob/2bf46a7728ff8fcf96374a882d2272d17595e311/setup.py) still reports Python 3.6 as supported, so I'm getting mixed signals here :sweat_smile:

kmuehlbauer:
Hi @1ace, it should work with 3.6, too. It's just that I've removed the 3.6 builds, which had temporary failures back then, which I can't reproduce now. 

But anyway, as we had to double the build matrix due to cover `h5py=2`/`h5py=3`, I would consider testing recent 3 Python versions as sufficient.

kmuehlbauer:
@1ace If you encounter any problems on your way, please open an issue with details.

1ace:
Awesome, thanks for the confirmation!

I'll make sure to report any issue when we do perform the upgrade :+1:

kmuehlbauer:
@1ace You might want to wait with your upgrade, see #84 and #85. While testing downstream (`xarray`) I've discovered a small but problematic inconsistency with the current code which lead to possible unwanted decoding of strings.

pikulmar:
@shoyer Are there any concrete plans for an updated (pre-)release with this fix included?

kmuehlbauer:
> Are there any concrete plans for an updated (pre-)release with this fix included?

Are you referring to this PR? This is actually available in `h5netcdf 0.9.0`. If you refer to #84, #85, no this is still worked out how and where this could potentially be tackled.

pikulmar:
> 
> 
> > Are there any concrete plans for an updated (pre-)release with this fix included?
> 
> Are you referring to this PR? This is actually available in `h5netcdf 0.9.0`. If you refer to #84, #85, no this is still worked out how and where this could potentially be tackled.

Ah, nice. I somehow only checked the GitHub releases, where 0.8.1. was still the latest. Thanks for replying so quickly and clarifying this!

kmuehlbauer:
@1ace @pikulmar 

FYI: It turned out, that in accordance with backwards compatibility with h5py2 and conformance with netcdf4 we had to restrict the string decoding to vlen strings only (#84/#85). So from h5netcdf 0.10.0 the keyword changed `decode_vlen_strings` and you can opt-in/opt-out early by setting to `True`/`False` (defaulting to `True`, will change to `False` in future versions).

Issue #84, opened 2021-02-09
h5py3 string encoding issues with xarray


kmuehlbauer:
While testing `xarray` with the new string encoding introduced in #82 I found that the current string_check is not sufficient and will result in unwanted decoding of data 

Currently the check is as follows:

https://github.com/h5netcdf/h5netcdf/blob/2bf46a7728ff8fcf96374a882d2272d17595e311/h5netcdf/core.py#L153-L155

This is a more specific test which only wraps strings with length=None with `.asstr()`.

```python
        if getattr(self._root, "decode_strings", False):
            string_info = h5py.check_string_dtype(self._h5ds.dtype)
            if string_info and string_info.length is None:
                return self._h5ds.asstr()[key]
```
While doing this I've found some more minor glitches (which are more or less style issue).

I've created two notebooks comparing current and proposed status https://gist.github.com/kmuehlbauer/f5b1c94b146d7041563527be0c0c84a6.

@shoyer I'll link a PR soon.

kmuehlbauer:
I've removed the `bug` - label since this might not actually be a bug in that sense, but some kind of interoperability issue.

kmuehlbauer:
This is a nice read of string handling in netcdf4-python https://unidata.github.io/netcdf4-python/#dealing-with-strings. And here similar for h5py https://docs.h5py.org/en/stable/strings.html and https://docs.h5py.org/en/stable/special.html

`xarray` uses `set_auto_chartostring(False)` to disable automatic conversion from char arrays to strings:

```python
    def get_array(self, needs_lock=True):
        ds = self.datastore._acquire(needs_lock)
        variable = ds.variables[self.variable_name]
        variable.set_auto_maskandscale(False)
        # only added in netCDF4-python v1.2.8
        with suppress(AttributeError):
            variable.set_auto_chartostring(False)
        return variable
```

@shoyer I think the current state of #85 is doing pretty much the same as `set_auto_chartostring` above. (This would still need some improvement.)

Let's take this h5dump (which is created via a numpy bytes array):

```
DATASET "bytes_var" {
      DATATYPE  H5T_STRING {
         STRSIZE 1;
         STRPAD H5T_STR_NULLTERM;
         CSET H5T_CSET_ASCII;
         CTYPE H5T_C_S1;
      }
      DATASPACE  SIMPLE { ( 6 ) / ( 6 ) }
      DATA {
      (0): "f", "o", "o", "b", "a", "r"
      }
[...]

PR #85, opened 2021-02-09
FIX: specifically check for vlen strings for string decoding

kmuehlbauer:
See #84 for in-depth analysis.

### DIFF
--- a/h5netcdf/core.py
+++ b/h5netcdf/core.py
@@ -558,7 +558,7 @@ class Group(Mapping):
                 if self._dim_sizes[dim] is None:
                     kwargs["maxshape"] = (None,)
                 self._h5group.create_dataset(
-                    name=dim, shape=(size,), dtype="S1", **kwargs
+                    name=dim, shape=(size,), dtype=">f4", **kwargs
                 )
 
             h5ds = self._h5group[dim]
--- a/h5netcdf/core.py
+++ b/h5netcdf/core.py
@@ -150,7 +150,7 @@ class BaseVariable(object):
         return self._h5ds.__array__(*args, **kwargs)
 
     def __getitem__(self, key):
-        if getattr(self._root, "decode_strings", False):
+        if getattr(self._root, "decode_vlen_strings", False):
             string_info = h5py.check_string_dtype(self._h5ds.dtype)
             if string_info and string_info.length is None:
                 return self._h5ds.asstr()[key]
@@ -689,7 +689,7 @@ class Group(Mapping):
 class File(Group):
     def __init__(self, path, mode="a", invalid_netcdf=None, phony_dims=None, **kwargs):
         if h5py.__version__ >= LooseVersion("3.0.0"):
-            self.decode_strings = kwargs.pop("decode_strings", None)
+            self.decode_vlen_strings = kwargs.pop("decode_vlen_strings", None)
         try:
             if isinstance(path, str):
                 if path.startswith(("http://", "https://", "hdf5://")):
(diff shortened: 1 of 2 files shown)


kmuehlbauer:
@shoyer So the base problem is that netcdf4 does not decode the `bytes_var` in the linked example notebook even though it is a string (but a fixed length string). The new behaviour especially tests for VLEN strings and only decodes those and let the fixed length strings as is. 

I've also found small differences in some of the special nectdf items (dimnsions etc.) and tried to move the current implementation to be as similar as possible. The only thing left is the NULLTERM vs. NULLPAD issue which I thin has no real (easy) solution.

kmuehlbauer:
Seems, that I'm pretty much lost in the `encoding` rabbit hole.  Decoding all strings might still be correct, though. I'm not really sure, where I'm thinking wrong, but that's obviously the case.... :disappointed: 

If there is a problem with 0.9.0 we might find out soon enough....

kmuehlbauer:
I'm closing this for now, since there is a solution ahead downstream.

kmuehlbauer:
@shoyer So, if we want to fix this issue here, we would have to apply this fix.

Unfortunately then `decode_strings` would need third meaning, "decode all but vlen char" or something like this.

h5netcdf/core.py
```diff
@@ -557,11 +558,11 @@ def _create_dim_scales(self):
                 if self._dim_sizes[dim] is None:
                     kwargs["maxshape"] = (None,)
                 self._h5group.create_dataset(
-                    name=dim, shape=(size,), dtype="S1", **kwargs
+                    name=dim, shape=(size,), dtype=">f4", **kwargs
```
shoyer, line 561: To clarify: these other dtype changes are unrelated to the vlen string stuff, right? I assume these are for consistency with netCDF?

shoyer:
> Unfortunately then `decode_strings` would need third meaning, "decode all but vlen char" or something like this.

Maybe we should just rename the option to `decode_vlen_strings`? That's all we need for compatibility with netCDF4.

Then I can issue another release. `decode_strings` hasn't been out in the world for long, so I think it's OK to remove it.

h5netcdf/core.py
kmuehlbauer: Yes, I tried to be as consistent with netcdf4 created files.

kmuehlbauer:
> Maybe we should just rename the option to `decode_vlen_strings`? That's all we need for compatibility with netCDF4.
> 
> Then I can issue another release. `decode_strings` hasn't been out in the world for long, so I think it's OK to remove it.

Would you consider having `decode_strings='vlen'` in addition to `decode_strings=True` and `decode_strings=False`? I'm just asking, because that way we would keep the possibility to decode all strings for other `h5netcdf` purposes.

shoyer:
> Would you consider having `decode_strings='vlen'` in addition to `decode_strings=True` and `decode_strings=False`? I'm just asking, because that way we would keep the possibility to decode all strings for other `h5netcdf` purposes.

We could do this, but netCDF's character type has a very weird representation in HDF5, restricted to strings with a fixed width of 1 bytes. As you noticed for Xarray, h5py's string decoding doesn't decode them properly.

We could potentially include a fixed version of this in h5netcdf (e.g., like netCDF4-python's auto-decoding of strings), but I don't think we'll ever want to apply h5py's string decoding directly in this case.

kmuehlbauer:
OK, then I'll rename `decode_strings` -> `decode_vlen_strings`. I'll push updates soon.

kmuehlbauer:
@shoyer I think this is good to go from my end.

PR #87, opened 2021-03-25
Include h5pyd.Dataset objects as netCDF variables

ajelenak:
A simple fix to include `h5pyd.Dataset` objects in the list of variables when scanning file content.

### DIFF
--- a/h5netcdf/core.py
+++ b/h5netcdf/core.py
@@ -19,9 +19,11 @@ try:
 except ImportError:
     no_h5pyd = True
     h5_group_types = (h5py.Group,)
+    h5_dataset_types = (h5py.Dataset,)
 else:
     no_h5pyd = False
     h5_group_types = (h5py.Group, h5pyd.Group)
+    h5_dataset_types = (h5py.Dataset, h5pyd.Dataset)
 
 __version__ = "0.10.0"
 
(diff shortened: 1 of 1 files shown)


kmuehlbauer:
@ajelenak Thanks. It looks like that's the last missing part of full `h5pyd` compatibility, right? 

We might consider setting up `h5pyd`-testing again. Do you know of any good resources for testing `h5pyd` in github actions CI?

ajelenak:
> It looks like that's the last missing part of full `h5pyd` compatibility, right?

Probably but I did not run the tests to verify.

> We might consider setting up `h5pyd`-testing again. Do you know of any good resources for testing with `h5pyd` in github actions CI?

This turned to be a pain the first time. I will ask around to see if we can improve this time. I agree it would good to have again.

kmuehlbauer:
> This turned to be a pain the first time. I will ask around to see if we can improve this time. I agree it would be good to have again.

I've skimmed the docs of `h5serv`. We could try to run `h5serv` on CI. But honestly, I've no idea how complicated this will be.

kmuehlbauer (approved):

kmuehlbauer:
Adding back testing of h5pyd based data should not hold off merge here. I'll merge by tomorrow. @ajelenak would we need to cut a release soon?

PR #92, opened 2021-04-16
update ``decode_vlen_strings`` FutureWarning

### LINKED ISSUE #90: Improve FutureWarning: String decoding changed with h5py >= 3.0.
I get many warnings like this one:

```
  /home/pierre/.pyenv/versions/3.8.2/lib/python3.8/site-packages/h5netcdf/core.py:765: FutureWarning: String decoding changed with h5py >= 3.0. Currently backwards compatibility with h5py < 3.0 is kept by decoding vlen strings per default. This will change in future versions for consistency with h5py >= 3.0. Setting 'decode_vlen_strings=True' forces string decoding.
    warnings.warn(msg, FutureWarning, stacklevel=0)
```

It seems to me that this warning could be improved. I don't know what I should do with it and how I should modify my code.

1. I guess some users (as me) don't understand what are "vlen strings". 
2. I don't know how and where I should change my code (the given line h5netcdf/core.py:765 is not useful for a user).
3. I get this warning with  h5py-3.2.1 and I don't mind about supporting h5py < 3.0.

I think an example of good warning message is

```
  /home/pierre/Dev/pythran/pythran/types/conversion.py:10: DeprecationWarning: `np.int` is a deprecated alias for the builtin `int`. To silence this warning, use `int` by itself. Doing this will not modify any behavior and is safe. When replacing `np.int`, you may wish to use e.g. `np.int64` or `np.int32` to specify the precision. If you wish to review your current use, check the release note link for additional information.
  Deprecated in NumPy 1.20; for more details and guidance: https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations
    numpy.int: 'npy_int',
```
The warning is emitted by Numpy but it shows exactly where is the problem (here, in Pythran). There is an good explanation (with a link) and it explains what has to be done to fix the issue. It would be great to improve the FutureWarning to get something as good! I guess it may not be as easy!

Just changing to `warnings.warn(msg, FutureWarning, stacklevel=2)` is already better since it points to the correct line in the code (where `h5netcdf.File` is used).

kmuehlbauer:
- closes #90

### DIFF
--- a/h5netcdf/core.py
+++ b/h5netcdf/core.py
@@ -759,10 +759,12 @@ class File(Group):
                 if self.decode_vlen_strings is None:
                     msg = (
                         "String decoding changed with h5py >= 3.0. "
+                        "See https://docs.h5py.org/en/latest/strings.html for more details. "
                         "Currently backwards compatibility with h5py < 3.0 is kept by "
                         "decoding vlen strings per default. This will change in future "
-                        "versions for consistency with h5py >= 3.0. "
-                        "Setting 'decode_vlen_strings=True' forces string decoding."
+                        "versions for consistency with h5py >= 3.0. To silence this "
+                        "warning set kwarg ``decode_vlen_strings=False``. Setting "
+                        "``decode_vlen_strings=True`` forces vlen string decoding."
                     )
                     warnings.warn(msg, FutureWarning, stacklevel=0)
                     self.decode_vlen_strings = True
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,7 @@ Development Version:
   By `Aleksandar Jelenak <https://github.com/ajelenak>`_.
 - Added automatic PyPI upload on creation of github release.
 - Moved Changelog to CHANGELOG.rst.
+- Updated ``decode_vlen_strings`` ``FutureWarning``
 
 Version 0.10.0 (February 11, 2021):
 


kmuehlbauer:
@paugier This would add the updated FutureWarning.

h5netcdf/core.py
```diff
@@ -757,10 +757,12 @@ def __init__(self, path, mode="a", invalid_netcdf=None, phony_dims=None, **kwarg
...
+                        "See https://docs.h5py.org/en/latest/strings.html for more details. "
                         "Currently backwards compatibility with h5py < 3.0 is kept by "
                         "decoding vlen strings per default. This will change in future "
-                        "versions for consistency with h5py >= 3.0. "
-                        "Setting 'decode_vlen_strings=True' forces string decoding."
+                        "versions for consistency with h5py >= 3.0. To silence this "
+                        "warning set kwarg ``decode_vlen_strings=False``. Setting "
+                        "``decode_vlen_strings=True`` forces vlen string decoding."
                     )
                     warnings.warn(msg, FutureWarning, stacklevel=0)
```
shoyer, line 765: Can you increase the stacklevel here? That will do a better job of pointing to user code that needs to be updated. Usually stacklevel=2 is the right call.

shoyer (approved):

h5netcdf/core.py
kmuehlbauer: Yes, I missed this one.

kmuehlbauer: Done.

kmuehlbauer:
I've rebased on latest master and added a changelog entry. Will merge, when CI is happy.

PR #99, opened 2021-04-20
Use 'r' for default file mode.

groutr:
I think the default file mode for h5netcdf should be 'r' instead of 'a'. This matches netCDF4 library (https://unidata.github.io/netcdf4-python/#Dataset.__init__) and the default mode for Python's open().

As a new user of h5netcdf, I was puzzled why the modification dates of my netCDF files were being updated when I was only opening the files for read operations as I assumed the default file mode would have been read-only. Opening a file in read-only mode is the least surprising way and, in this case, the least destructive way to open netCDF file.

### DIFF
--- a/h5netcdf/tests/test_h5netcdf.py
+++ b/h5netcdf/tests/test_h5netcdf.py
@@ -212,7 +212,7 @@ def write_h5netcdf(tmp_netcdf):
 
 
 def read_legacy_netcdf(tmp_netcdf, read_module, write_module):
-    ds = read_module.Dataset(tmp_netcdf)
+    ds = read_module.Dataset(tmp_netcdf, "r")
     assert ds.ncattrs() == ["global", "other_attr"]
     assert ds.getncattr("global") == 42
     if write_module is not netCDF4:
@@ -263,7 +263,7 @@ def read_legacy_netcdf(tmp_netcdf, read_module, write_module):
     # Check the behavior if h5py. Cannot expect h5netcdf to overcome these
     # errors:
     if is_h5py_char_working(tmp_netcdf, "z"):
-        ds = read_module.Dataset(tmp_netcdf)
+        ds = read_module.Dataset(tmp_netcdf, "r")
         v = ds.variables["z"]
         assert array_equal(v, _char_array)
         assert v.dtype == "S1"
@@ -272,7 +272,7 @@ def read_legacy_netcdf(tmp_netcdf, read_module, write_module):
         assert v.ncattrs() == ["_FillValue"]
         assert v.getncattr("_FillValue") == b"X"
     else:
-        ds = read_module.Dataset(tmp_netcdf)
+        ds = read_module.Dataset(tmp_netcdf, "r")
 
     v = ds.variables["scalar"]
     assert array_equal(v, np.array(2.0))
@@ -309,7 +309,7 @@ def read_legacy_netcdf(tmp_netcdf, read_module, write_module):
 
 def read_h5netcdf(tmp_netcdf, write_module, decode_vlen_strings):
     remote_file = isinstance(tmp_netcdf, str) and tmp_netcdf.startswith(remote_h5)
-    ds = h5netcdf.File(tmp_netcdf, **decode_vlen_strings)
+    ds = h5netcdf.File(tmp_netcdf, "r", **decode_vlen_strings)
     assert ds.name == "/"
     assert list(ds.attrs) == ["global", "other_attr"]
     assert ds.attrs["global"] == 42
@@ -356,7 +356,7 @@ def read_h5netcdf(tmp_netcdf, write_module, decode_vlen_strings):
     ds.close()
 
     if is_h5py_char_working(tmp_netcdf, "z"):
-        ds = h5netcdf.File(tmp_netcdf)
+        ds = h5netcdf.File(tmp_netcdf, "r")
         v = ds["z"]
         assert array_equal(v, _char_array)
         assert v.dtype == "S1"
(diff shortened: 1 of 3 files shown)


kmuehlbauer:
@groutr The mode='a' has been there from the beginning, @shoyer might have ideas why this is used.

It seems reasonable to switch to mode='r' as default (h5py uses same default, too).

The test suite is obviously relying on that feature and would need to be changed accordingly.

shoyer:
I think I just blindly copied h5py, which used to default to `mode='a'`. I agree that read-only mode is a better default.

kmuehlbauer:
@groutr To make the tests work, the file open `mode` would have to fixed. I've taken the chance to do the search/replace. You can grab from my branch https://github.com/kmuehlbauer/h5netcdf/commit/17bfc283372357b824cf64bb19836d2c5b73b9d5, or I can push to your branch.

groutr:
@kmuehlbauer Thanks! I pulled your branch into my PR.

shoyer:
It would be nice to do this change in way that doesn't break existing users of h5netcdf immediately by going through a deprecation cycle.

This would mean switching the default to `mode=None` and issuing a `FutureWarning` about the changed default if a user hasn't specified `mode` explicitly. Then in the future we can safely switch to `mode='r'`, after users have had some warning about it.

shoyer:
> * we use `mode=None`
> * if `mode=None` we issue FutureWarning and set `mode='a'`
> * in one of the next versions we use `mode='r'`

Yes, this is exactly what I had in mind

kmuehlbauer:
@groutr The approach suggested by @shoyer is the way to go. Unfortunately this will need more treatment to the code. Please let me know, if you can spare the cycles to get the PR along. If not I can take a look to get this in soon.

groutr:
I will try to get to it tonight.

kmuehlbauer:
Thanks @groutr, this is almost there. I've added another commit on top of your latest changes to my branch (https://github.com/kmuehlbauer/h5netcdf/commit/9d87e6f08e8b4a45bd072ea9dd69e285a326f058) which fixes the testsuite and style. Also added a test.

groutr:
@kmuehlbauer Thank you for the assistance. I was having some trouble getting the tests to execute on my local machine. I've got it sorted out now.

groutr:
@kmuehlbauer could you please explain the `stacklevel=2` argument to warnings?

groutr:
> I think I just blindly copied h5py, which used to default to `mode='a'`. I agree that read-only mode is a better default.

The documentation to h5py now includes this note:

> Changed in version 3.0: Files are now opened read-only by default. Earlier versions of h5py would pick different modes
> depending on the presence and permissions of the file.

kmuehlbauer:
> could you please explain the stacklevel=2 argument to warnings? (ie why 2?)

It shows the warning with the location of the code which need to be changed, not where it breaks.

kmuehlbauer:
@groutr GitHub added a security feature. PR's from forks have to be approved before running the test suite. Need to look how to get a ping on this.

h5netcdf/tests/test_h5netcdf.py
```diff
@@ -574,6 +574,12 @@ def test_decode_string_error(tmp_local_or_remote_netcdf):
             assert ds.name == "/"
 
 
+def test_mode_warning(tmp_local_or_remote_netcdf):
+    with pytest.warns(FutureWarning):
+        with h5netcdf.File(tmp_local_or_remote_netcdf) as ds:
+            pass
```
kmuehlbauer, line 580: Seems `flake8` doesn't like that `ds` isn't used. We could `assert ds` instead of `pass`.

groutr: What if we did
```python
with pytest.warns(FutureWarning):
    h5netcdf.File(tmp_local_or_remote_netcdf).close()
```

kmuehlbauer: Let's just duplicate the code of the two tests above. But I have no strong feelings what to choose. It should just work.

kmuehlbauer (approved):
@groutr Thanks a lot. I'll add an item CHANGELOG.rst acknowledging your contribution, if that's OK with you.

groutr:
@kmuehlbauer That's fine with me. Thank you for help and guidance in this PR.

kmuehlbauer:
@shoyer Could you have one last look, before merge?

h5netcdf/tests/test_h5netcdf.py
```diff
@@ -108,7 +108,7 @@ def array_equal(a, b):
 def is_h5py_char_working(tmp_netcdf, name):
     h5 = get_hdf5_module(tmp_netcdf)
     # https://github.com/Unidata/netcdf-c/issues/298
-    with h5.File(tmp_netcdf, "r") as ds:
+    with h5.File(tmp_netcdf) as ds:
```
shoyer, line 111: This should probably still explicitly set `mode='r'`, unless you are writing a test explicitly to check for the warning message -- which I'm pretty sure is not the case here.

h5netcdf/tests/test_h5netcdf.py
```diff
@@ -788,6 +794,7 @@ def test_failed_read_open_and_clean_delete(tmpdir):
     # A file that does not exist but is opened for
     # reading should only raise an IOError and
     # no AttributeError at garbage collection.
+    # gc.collect()
```
shoyer, line 797: delete?

shoyer (approved):

h5netcdf/tests/test_h5netcdf.py
kmuehlbauer: Sure, added. I've added an explicit test for the FutureWarning too.

h5netcdf/tests/test_h5netcdf.py
kmuehlbauer: Thanks!

kmuehlbauer:
Thanks @groutr!

PR #101, opened 2021-04-23
WIP: refactor dimension scales handling

### LINKED ISSUE #88: dimension processing performance when appending to file
As reported by @aldanor in https://github.com/pydata/xarray/issues/2857 `h5netcdf` shows performance degradation when consecutively appending groups with datasets having dimensions.

The problem is that for every append the whole file/group/subgroup structure is traversed to find unassigned dimensions, create the missing dimension scales and attach it properly to the file.

A quick check commenting the three lines of code handling the dimensions:

https://github.com/h5netcdf/h5netcdf/blob/0d6c2268fda93bd0a9e98a3e1a8e305f011df5a2/h5netcdf/core.py#L857-L863

shows, that this resolves the performance hit (but not writing any dimension scales).

A possible solution was suggested by @shoyer https://github.com/pydata/xarray/issues/2857#issuecomment-807319582:

> I suspect this could be solved by adding an optimization into h5netcdf to only call _attach_dim_scales() (and maybe some other methods) on variables/groups that have been modified (as opposed to the entire file).

Please see also discussion including extensive tests provided by @aldanor in https://github.com/pydata/xarray/issues/2857.

kmuehlbauer:
Fixes #88 

UPDATE: This PR was split into several smaller ones, it will be kept here for reference until completely resolved.

- Dimension scales are now created and attached on the fly, instead of doing the iteration on `flush` over all groups/subgroups.
- When opening a file for writing the maximum dimension id (`_Netcdf4Dimid`) is determined and increased for every newly created dimension (#104)
- Transparent handling of `_nc4_non_coord` (#102)
- Refactored out several functions for dim_scale handling at `h5netcdf.Group` and `h5netcdf.Variable` level
- Automatic resizing of variables and attached unlimited dimensions when writing to variables (#103)
- Adapted `resize_dimension` for both dimension resizing and recursive/non-recursive variable resizing (#103)
- Refactored out `_create_h5netcdf_variable`-function which creates hdf5 dataset and `h5netcdf.Variable` (#105)
- Simplified `_create_child_variable`-function with added checks which (netcdf) variable type will be created (#105) 

Bullet point 2 is now the last bottleneck since we need to iterate over all datasets to get the dimension id. I've used `visititems()` for the iteration.  There is room for further refactoring.

Todo:
- [x] simplifiy variable type checks
- [x] add/update tests for resizing dimensions/variables
- [x] add tests for dimension scale handling (eg. attaching/detaching dimension scales)
- [x] add tests for _nc4_non_coord_ handling
- [x] add/update tests for variable creation
- [ ] add more tests

For performance testing I used the code provided by @aldanor:
```python
import time
import xarray as xr
import numpy as np
import h5py

arr = xr.DataArray(np.random.RandomState(0).randint(-100, 100, (50_000, 3)), dims=['x', 'y'])
ds = xr.Dataset({'arr': arr})

filename = 'test.h5'
save = lambda group: ds.to_netcdf(filename, engine='h5netcdf', mode='a', group=str(group))

with h5py.File(filename, 'w') as _:
    pass

for i in range(250):
    save(i)
```
|  netcdf4    |  h5netcdf 0.11.0  |  h5netcdf-refactor |
|-----|------|-----|
| 15 sec   | 208 sec | 20 sec |

So this is a great improvement, although we do not match `netCDF4`.

Some valuable links on dimension scales:
- https://www.unidata.ucar.edu/blogs/developer/en/entry/dimensions_scales
- https://www.unidata.ucar.edu/blogs/developer/en/entry/dimension_scale2
- https://www.unidata.ucar.edu/blogs/developer/en/entry/dimension_scales_part_3
- https://www.unidata.ucar.edu/blogs/developer/en/entry/netcdf4_shared_dimensions
- https://www.unidata.ucar.edu/blogs/developer/en/entry/netcdf4_use_of_dimension_scales
- https://www.researchgate.net/publication/330347054_2A5_NETCDF-4_PERFORMANCE_IMPROVEMENTS_OPENING_COMPLEX_DATA_FILES

String `NULLTERM` vs `NULLPAD`:
- https://github.com/PyTables/PyTables/issues/264

Testing Notebook:
- https://gist.github.com/kmuehlbauer/99ad4749c4e5923b1364187c6363930a

netcdf4-python quirks:
- `_Netcdf4Dimid` gets attached to all data variables if a 2D coordinate variable is created  and any variable is written/file is reopened for append, see https://github.com/Unidata/netcdf4-python/issues/1104
- unlimited variable dimensions are reported as current size of the dimension scale, even if the variable's underlying `DATASPACE` dimension is smaller (eg.0)

### DIFF
--- a/h5netcdf/core.py
+++ b/h5netcdf/core.py
@@ -104,51 +104,30 @@ def _expanded_indexer(key, ndim):
     return tuple(new_key)
 
 
-def _check_coord_var(grp, name, dimensions):
-    """Check for coordinate variable"""
-    name_in_dims = name in grp.dimensions
-    name_in_var_dims = name in dimensions
-    # 1d coordinate
+def _check_netcdf4_variable_type(grp, name, dimensions):
+    """Return tuple with variable types."""
+    grp_dims = name in grp.dimensions
+    var_dims = name in dimensions
+
     _1d = len(dimensions) == 1
-    # 2d coordinate
     _2d = (
         len(dimensions) == 2
         and name == dimensions[0]
         and dimensions[1] not in grp.variables
         and dimensions[1] in grp.dimensions
     )
-    return name_in_dims and name_in_var_dims and (_1d or _2d)
 
-
-def _check_non_coord_var(grp, name, dimensions):
-    """Check for _nc4_non_coord_"""
-    name_in_dims = name in grp.dimensions
-    name_in_var_dims = name in dimensions
-    return (
-        (name_in_dims and not name_in_var_dims)
-        or (name_in_dims and name_in_var_dims and len(dimensions) > 1)
-    ) and (not _check_coord_var(grp, name, dimensions))
-
-
-def _check_data_var(grp, name, dimensions):
-    """Check for data variable"""
-    name_in_dims = name in grp.dimensions
-    name_in_var_dims = name in dimensions
-    return (
-        (name_in_dims and not name_in_var_dims)
-        or (name_in_dims and name_in_var_dims and len(dimensions) > 1)
-        or (not name_in_dims and not name_in_var_dims)
-    ) and (not _check_coord_var(grp, name, dimensions))
-
-
-def _check_var_with_coords(grp, name, dimensions):
-    """Check for variable with coordinates"""
-    name_in_dims = name in grp.dimensions
-    name_in_var_dims = name in dimensions
-    return (
-        (not name_in_dims and not name_in_var_dims)
-        or (name_in_dims and name_in_var_dims and len(dimensions) > 1)
-    ) and (not _check_coord_var(grp, name, dimensions))
+    nc4_var_type = ()
+    # either coord variable
+    if grp_dims and var_dims and (_1d or _2d):
+        nc4_var_type += ("coord",)
+    # or data variable
+    else:
+        if not var_dims or (var_dims and len(dimensions) > 1):
+            nc4_var_type += ("data",)
+            if grp_dims:
+                nc4_var_type += ("non_coord",)
+    return nc4_var_type
 
 
 class BaseVariable(object):
@@ -650,12 +629,15 @@ class Group(Mapping):
                 if d not in self.dimensions:
                     self.dimensions[d] = s
 
+        # get netcdf4 variable types
+        nc4_var_type = _check_netcdf4_variable_type(self, name, dimensions)
         refs = None
-        if _check_coord_var(self, name, dimensions):
+
+        if "coord" in nc4_var_type:
             refs = self._get_dim_scale_refs(name)
             self._delete_dim_scale(name)
 
-        if _check_non_coord_var(self, name, dimensions):
+        if "non_coord" in nc4_var_type:
             name = "_nc4_non_coord_" + name
 
         variable = self._create_h5netcdf_variable(
@@ -663,19 +645,17 @@ class Group(Mapping):
         )
 
         # This is a bit of a hack, netCDF4 attaches _Netcdf4Dimid to every variable
-        # in the moment when a variable is first written to, after variable creation.
+        # when a variable is first written to, after variable creation.
         # Here we just attach it to every variable on creation.
         variable._ensure_dim_id()
 
-        if _check_coord_var(self, name, dimensions):
+        if "coord" in nc4_var_type:
             self._create_dim_scale(name)
             if refs is not None:
                 self._attach_dim_scale(name, refs)
 
-        if _check_data_var(self, name, dimensions):
+        if "data" in nc4_var_type:
             variable._attach_dim_scales()
-
-        if _check_var_with_coords(self, name, dimensions):
             variable._attach_coords()
 
         return variable
--- a/h5netcdf/core.py
+++ b/h5netcdf/core.py
@@ -623,11 +623,6 @@ class Group(Mapping):
         self._variables[name] = self._variable_cls(self, name, dimensions)
         variable = self._variables[name]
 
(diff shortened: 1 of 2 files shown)


h5netcdf/core.py
```diff
@@ -726,6 +747,7 @@ def __init__(self, path, mode="a", invalid_netcdf=None, phony_dims=None, **kwarg
             self._closed = False
 
         self._mode = mode
+        self._writable = not (mode == "r")
```
groutr, line 750: Would this be the same as `mode != 'r'`?

kmuehlbauer: Now that you mention this...This looks much nicer. Why did I do this in the first place?

h5netcdf/core.py
```diff
@@ -674,6 +690,11 @@ def resize_dimension(self, dimension, size):
         # Resize the dimension.
         self._current_dim_sizes[dimension] = size
 
+        # If not a variable but dimension -> resize
+        if dimension not in self.variables and dimension in self._h5group:
+            self._h5group[dimension].resize((size,))
```
shoyer, line 695: This looks sensible, but is it related somehow to the other changes? It might warrant another unit-test.

shoyer (commented):
what a nice optimization! 😍

h5netcdf/core.py
kmuehlbauer: If only a dimension is created, the corresponding Dataset will be created at the same time. Before it would only be created on `flush`.

Now, when we create a dimension and resize it afterwards, we have to resize here since the Dataset already exists. 

I'll take a look and come up with some more tests with this. Good chance that this will lead to some more refactoring.

kmuehlbauer:
OK; I've found some inconsistencies with the current code, the refactored code and the netcdf behaviour. I'll need to dive a bit deeper.

kmuehlbauer:
@shoyer I've tried hard to break this into pieces, but it's too much functionally interdepending.

I've worked on these things:

- transparent handling of `_nc4_non_coord`
- automatic resizing of variables and attached unlimited dimensions when writing to variables
- refactored out `_create_h5netcdf_variable`-function which creates hdf5 dataset and `h5netcdf.Variable`
- simplified `_create_child_variable`-function with added checks which (netcdf) variable type will be created (these checks need another refactor, since the boolean logic hard to read atm)
- refactored out several functions for dim_scale handling at `h5netcdf.Group` and `h5netcdf.Variable` level
- dim_scales are created/removed and attached/detached on the fly
- adapted `resize_dimension` for both dimension resizing and recursive/non-recursive variable resizing

I did not touch the tests yet, since I wanted to be fully backwards compatible regarding existing tests. That all tests run is good indicator that the refactor was successful. I've also checked the details (netCDF4, h5netcdf, legacyapi) inside a jupyter notebook using kind of read/write matrix:

- write netCDF4, write legacyapi, write h5netcdf
- read netCDF4, read legacyapi, read h5netcdf
- checking `ncdump` and `h5dump` on all created files, diffing the output to check for consistency

I'll make the notebook available, after a clean-up. Tests follow too. Looking forward to hear your thoughts.

kmuehlbauer:
@shoyer I've added a link to the testing notebook in the OP.

h5netcdf/core.py
kmuehlbauer: @shoyer I've added tests, which checks correct resizing of dimensions vs. variables.

kmuehlbauer:
@shoyer This is ready from my side for review.

shoyer:
With regards to the test coverage -- right now, it looks like you are mostly testing h5netcdf only private APIs. This is a fine start, but would be ideal to also test public interfaces, and to verify that netCDF4 creates equivalent HDF5 files.

h5netcdf/core.py
```diff
@@ -73,6 +72,66 @@ def _invalid_netcdf_feature(feature, allow, file, stacklevel=0):
...
+def _expanded_indexer(key, ndim):
+    """Given a key for indexing an ndarray, return an equivalent key which is a
+    tuple with length equal to the number of dimensions.
+
+    The expansion is done by replacing all `Ellipsis` items with the right
+    number of full slices and then padding the key with full slices so that it
+    reaches the appropriate dimensionality.
+    """
+    # Taken from xarray:
+    # https://github.com/pydata/xarray/blob/835a53e62b1e5018faa323c649149e8294dd6af7/xarray/core/indexing.py#L23
```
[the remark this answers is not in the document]
kmuehlbauer: @shoyer Thanks, I'll add it.

kmuehlbauer:
So the round-trip tests are not sufficient? What do you suggest to test this?

kmuehlbauer:
I can take the example from the Unidata blog and enhance it a bit.

shoyer:
> So the round-trip tests are not sufficient? What do you suggest to test this?

Well, we do still have the existing checks. And if this change is only speeding things up (not changing functionality), then they are probably fine.

kmuehlbauer:
@shoyer It changes functionality somehow in that sense that every change to dimensions will instantly be applied to the underlying datasets and not on `close()`/`flush()`. That has effects when variables are created and written. I'm trying to address everything but it can't be split into pieces because of the involved machinery. With every step forward I encounter another inconsistency between current h5netcdf behaviour and netCDF4.

I'll make this WIP again. This needs much more love.

kmuehlbauer:
@shoyer One example of those inconsistencies: https://github.com/Unidata/netcdf4-python/issues/1104

kmuehlbauer:
This PR has been superseded by #102, #103, #104 and #112. I'll close this. Refernces in the first post will be moved to the Wiki.

PR #102, opened 2021-05-11
handle _nc4_non_coord_ 

kmuehlbauer:
This moves the handling of `_nc4_non_coord_` variables into `BaseVariable`.

needed for #101


kmuehlbauer:
@shoyer I'm going to strip everything from #101 which can be separated easily. Hopefully this will make reviewing easier.

shoyer (commented):
Would it make sense to add any new tests here?

PR #103, opened 2021-05-11
refactor resize dimensions

kmuehlbauer:
- resize dimensions/variables when writing to `Variables` (as netcdf4-python does) only for legacy API
- return padded arrays (up to current size of unlimited dimensions) for both API's
- refactor `resize_dimensions`
- enhanced tests

Update: changed list (#101 is not needed anymore)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -48,7 +48,7 @@ Usage
 
 h5netcdf has two APIs, a new API and a legacy API. Both interfaces currently
 reproduce most of the features of the netCDF interface, with the notable
-exception of support for operations the rename or delete existing objects.
+exception of support for operations that rename or delete existing objects.
 We simply haven't gotten around to implementing this yet. Patches
 would be very welcome.
 
@@ -90,6 +90,14 @@ design is an adaptation of h5py to the netCDF data model. For example:
         # explicitly resize a dimension and all variables using it
         f.resize_dimension('z', 3)
 
+Note:
+
+- Automatic resizing of unlimited dimensions with array indexing is not available.
+- Dimensions need to be manually resized with ``Group.resize_dimension(dimension, size)``.
+- Arrays are returned padded with ``fillvalue`` (taken from underlying hdf5 dataset) up to
+  current size of variable's dimensions. The behaviour is equivalent to netCDF4-python's
+  ``Dataset.set_auto_mask(False)``.
+
 Legacy API
 ~~~~~~~~~~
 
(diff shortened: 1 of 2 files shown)


shoyer:
> The difference is that `netcdf4-python`/`netcdf-c` changes the dimension size but not the size of all connected variables (in the underlying DATASPACE). Instead it just reports the Variable's having the size of its dimensions. On access it will return a masked array filling not available data. 

This seems like really messy logic to me, but I guess we could copy it. Presumably it should be able to make a netcdf file with only some expanded variables with netcdf4-python, which we could compare against for testing.

I think it's much more important to **read** data like this than to **write** it. I'm not even sure I would support writing expanded variables via assignment at all in the non-legacy API. In general, explicitly enlarging dimensions seems like a better approach to me.

kmuehlbauer:
Yes, there are some quirks in `netcdf4-python`/`netcdf-c` as well as in `hdf5`. 

One reason for such behaviour might be if you create hundreds of variables with an unlimited dimension and then fill only one variable up to a reasonable large size, the other variables are not affected on disk (saving space). Not sure if this was the intention in the first place, but it does make some sense at least.

I did not add the reading part yet, since we would need to have access to the dimension scales on read time. This would work for the `mode='r'` case when dimension scales are available, but while creating dimensions/variables this won't work since the dimension scales are currently created at file close. It's very much interconnected everything.

If I got all right, then I'll move forward with this PR like this:

- apply the `netcdf4-python` behaviour to the `legacyapi` on read/write
- apply the `netcdf4-python` behaviour to the new api only for read (and force the user to enlarge dimension before write)
- create special test case using `mode="r"` to test reading against `netcdf4-python` testfile

One more question: How should the `resize_dimension` work? Should it resize all connected variables (as it does now), or should it just resize the dimension?

kmuehlbauer:
@shoyer I've added the necessary changes according to your suggestions. Also updated `README.txt`.

h5netcdf/core.py
```diff
@@ -157,9 +230,28 @@ def __getitem__(self, key):
             string_info = h5py.check_string_dtype(self._h5ds.dtype)
             if string_info and string_info.length is None:
                 return self._h5ds.asstr()[key]
+
+        # return array padded with fillvalue
+        if self.dtype != str and self.dtype.kind in ["f", "i", "u"]:
```
kmuehlbauer, line 235: To things are currently unclear to me:

- is `["f", "i", "u"]` enough for now?
- is `np.pad` the right function for this?

kmuehlbauer:
Rebased on latest master here, too.

kmuehlbauer:
Found another glitch while working on #101. This is now ready for review.

kmuehlbauer:
@shoyer I'll come back to this after #104 and #112 are in.

kmuehlbauer:
rebased on latest master

kmuehlbauer:
another rebase after #104 and #112 are in

kmuehlbauer:
Update: 

After the dimension scale refactor the current status of this PR is as follows:

- resize dimensions/variable when writing to a `variable` (as netcdf4-python does) only for legacy API, but do not resize other variables (which share affected unlimited dimensions)

- return padded arrays (with fillvalue, up to current size of unlimited dimensions) for both API's

- report sizes from connected dimensions, not underlying h5 datasets for both API's

@shoyer Before I clean-up things here, it would be great if you could have a another look at this.

kmuehlbauer:
I've combined and rebased commits on latest master. This is ready for review from my side.

README.rst
```diff
@@ -90,6 +90,14 @@ design is an adaptation of h5py to the netCDF data model. For example:
         # explicitly resize a dimension and all variables using it
         f.resize_dimension('z', 3)
 
+Pitfalls of the new API:
```
kmuehlbauer, line 93: Maybe just use "Note:" instead if "Pitfalls..."?

kmuehlbauer:
@shoyer It would be great, if you could review this PR. It's one of the big chunks (together with #127) before we can move to h5netcdf 1.0.

kmuehlbauer:
Rebased on latest master again.

PR #107, opened 2021-05-20
write _NCProperties when overwriting file (mode="w")

### LINKED ISSUE #111: _NCProperties isn't written when overwriting existing files
When overwriting existing files (`mode="w"`) the `_NCProperties` don't get written.

kmuehlbauer:
This fixes an issue where `_NCProperties` isn't written when overwriting existing files.

- added a test which checks if attribute gets written

closes #111


kmuehlbauer:
@shoyer Sorry for the many PullRequests targeting different aspects. My suggestion would be to revise every PR, ping you for review and merge one after the other. So this would be the first to tackle.

shoyer (approved):
Sounds like a good plan! This one looks good to me.

By the way, I do appreciate your enthusiasm here!

kmuehlbauer:
> Sounds like a good plan! This one looks good to me.

Great!

> By the way, I do appreciate your enthusiasm here!

:+1:

PR #108, opened 2021-05-20
Create/Attach dimension scales on append (``mode="r+"``)

kmuehlbauer:
This fixes an issue where dimension scales weren't created/attached when appending with ``mode="r+"``.

- added tests to check for scales when appending with `mode="r+"`


kmuehlbauer:
@shoyer Here also fixing a minor but problematic issue.

kmuehlbauer:
@shoyer This is ready for review/merge. As this change is really very simple, I'll merge by tomorrow. We can adapt later if necessary.

shoyer (approved):

Issue #106, opened 2021-05-18
Creating variable colliding with dimension name breaks


kmuehlbauer:
This breaks with latest master as well as with v 0.11.0. Reason is trying to detach dimension scale (`x`) from variable (`z`) which hasn't been attached yet. It would have been attached on flush/close.

```python
with h5netcdf.File("test.nc", "w") as f:
    f.dimensions["x"] = 2
    f.dimensions["y"] = 2

with h5netcdf.File("test.nc", "a") as f:
    f.create_variable("z", dimensions=("x", ), dtype=np.int64)
    f.create_variable("x", dimensions=("y", ), dtype=np.int64) 
```

```python
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-8-73def0570bc0> in <module>
      5 with h5netcdf.File("test.nc", "a") as f:
      6     f.create_variable("z", dimensions=("x", ), dtype=np.int64)
----> 7     f.create_variable("x", dimensions=("y", ), dtype=np.int64)

/home/data/kai/projects/h5netcdf/h5netcdf/core.py in create_variable(self, name, dimensions, dtype, data, fillvalue, **kwargs)
    526             group = group._require_child_group(k)
    527         return group._create_child_variable(
--> 528             keys[-1], dimensions, dtype, data, fillvalue, **kwargs
    529         )
    530 

/home/data/kai/projects/h5netcdf/h5netcdf/core.py in _create_child_variable(self, name, dimensions, dtype, data, fillvalue, **kwargs)
    500             if _netcdf_dimension_but_not_variable(h5ds):
    501                 self._detach_dim_scale(name)
--> 502                 del self._h5group[name]
    503 
    504         self._h5group.create_dataset(

/home/data/kai/projects/h5netcdf/h5netcdf/core.py in _detach_dim_scale(self, name)
    608                     var._h5ds.dims[n].detach_scale(self._all_h5groups[dim])
    609 
--> 610         for subgroup in self.groups.values():
    611             if dim not in subgroup._h5group:
    612                 subgroup._detach_dim_scale(name)

~/miniconda3/envs/h5py3/lib/python3.9/site-packages/h5py/_hl/dims.py in detach_scale(self, dset)
    100         """
    101         with phil:
--> 102             h5ds.detach_scale(self._id, dset.id, self._dimension)
    103 
    104     def items(self):

h5py/_objects.pyx in h5py._objects.with_phil.wrapper()

[...]

PR #109, opened 2021-05-20
create/attach/detach dimension scales only if necessary

kmuehlbauer:
This fixes the issue in #106. 

- create and attach scales only once
- added tests
- correct handling of special n-dimensional `_nc4_non_coord_` variable (name clashing)

### DIFF
--- a/h5netcdf/tests/test_h5netcdf.py
+++ b/h5netcdf/tests/test_h5netcdf.py
@@ -1159,3 +1159,59 @@ def test_scales_on_append(tmp_local_netcdf):
     # check scales
     with h5netcdf.File(tmp_local_netcdf, "r") as ds:
         assert ds.variables["test1"].attrs._h5attrs.get("DIMENSION_LIST", False)
+
+
+def test_create_attach_scales(tmp_local_netcdf):
+    # create file with netCDF4
+    with netCDF4.Dataset(tmp_local_netcdf, "w") as ds:
+        ds.createDimension("x", 0)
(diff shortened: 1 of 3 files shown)


kmuehlbauer:
@shoyer This too is ready for review/merge.

h5netcdf/tests/test_h5netcdf.py
```diff
@@ -1137,3 +1137,63 @@ def test_overwrite_existing_file(tmp_local_netcdf):
...
+
+    # check scales
+    with h5netcdf.File(tmp_local_netcdf, "r") as ds:
+        refs = ds._h5group["x"].attrs.get("REFERENCE_LIST", False)
+        assert len(refs) == 3
+        for (ref, dim), name in zip(refs, ["/test", "/test1", "/_nc4_non_coord_y"]):
+            assert dim == 0
+            assert ds._root._h5file[ref].name == name
+
+    # create file with netCDF4
```
shoyer, line 1163: These lines look duplicated from above?

shoyer (approved):

h5netcdf/tests/test_h5netcdf.py
kmuehlbauer: Yes, correct, I recreate the file to append with legacyapi. I could create a special function to reduce the duplication.

kmuehlbauer: I've de-duplicated the test code. Thanks @shoyer.

PR #110, opened 2021-05-20
Switch warning into error when using invalid netCDF features

### LINKED ISSUE #100: Switch using invalid netCDF features from a warning into an error?
We added a warning when writing invalid netCDF files back in 2017:
https://github.com/h5netcdf/h5netcdf/pull/29

This was intended as an incremental step to preserve backwards compatibility. At this point, it would _probably_ be better to switch h5netcdf to raise an error in these cases instead. Or if we don't actually want to make that change, we should remove the note about the warning turning into an error changing in the future :)

kmuehlbauer:
Resolves #100

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,6 +13,8 @@ Development Version:
   By `Kai Mühlbauer <https://github.com/kmuehlbauer>`_.
 - Create/Attach/Detach dimension scales only if necessary.
   By `Kai Mühlbauer <https://github.com/kmuehlbauer>`_.
+- Switch warning into error when using invalid netCDF features.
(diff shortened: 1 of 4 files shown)


kmuehlbauer:
Updated `README.rst` and removed stale comments. Should be ready for review.

kmuehlbauer:
@shoyer I'll rebase this when #108 and #109 are in, but you might have a look already.

kmuehlbauer:
I've rebased on latest master, @shoyer. This is ready for review.

shoyer (approved):
I wonder, would this be a good time to also declare version 1.0?

PR #104, opened 2021-05-11
assign dimensions on creation

kmuehlbauer:
This retrieves the `max_dim_id` count on start if the file is writable and assigns the dimids on the fly.

Update: This changes the current behaviour of iterating over the groups/datasets at file close to iteration at file start. `dim_order` is assigned on the fly instead at file close.This is needed to create/attach and detach/delete dimension scales on the fly (in follow-up PR). 

needed for #101

PR #112, opened 2021-06-10
Refactor dimension scale handling

### LINKED ISSUE #88: dimension processing performance when appending to file

kmuehlbauer:
This PullRequest refactors the dimension scale handling from creating/attaching at file close to an on-the-fly creation/attaching process. This improves performance drastically (see #101 for example).

The example added builds on the nice example at the [unidata blog ](https://www.unidata.ucar.edu/blogs/developer/en/entry/netcdf4_shared_dimensions).

- builds on #104
- replaces #101 (at least partly)
- closes #88

I'll rebase when #104 is merged.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,6 +5,9 @@ Development Version:
 
 - Assign dimensions at creation time, instead of at sync/flush (file-close).
   By `Kai Mühlbauer <https://github.com/kmuehlbauer>`_.
+- Create/attach dimension scales on the fly, instead of at sync/flush (file-close).
+  By `Kai Mühlbauer <https://github.com/kmuehlbauer>`_.
+
 
 Version 0.12.0 (December 20, 2021):
 
--- a/h5netcdf/core.py
+++ b/h5netcdf/core.py
@@ -99,7 +99,8 @@ class BaseVariable(object):
                 order_dim[coord_id] for coord_id in attrs["_Netcdf4Coordinates"]
             )
 
-        child_name = self.name.split("/")[-1]
+        # need to use the h5ds name here to distinguish from collision dimensions
+        child_name = self._h5ds.name.split("/")[-1]
         if child_name in self._parent.dimensions:
             return (child_name,)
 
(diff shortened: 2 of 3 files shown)


kmuehlbauer:
@shoyer This one builds on #104. If #104 gets merged, I can rebase here. Or we can decide to skip #104 and merge this one instead.

kmuehlbauer:
@shoyer This is ready for review, too.

kmuehlbauer:
Latest perormance check with the code provided by @aldanor:

```python
import time
import xarray as xr
import numpy as np
import h5py

arr = xr.DataArray(np.random.RandomState(0).randint(-100, 100, (50_000, 3)), dims=['x', 'y'])
ds = xr.Dataset({'arr': arr})

filename = 'test.h5'
save = lambda group: ds.to_netcdf(filename, engine='h5netcdf', mode='a', group=str(group))

with h5py.File(filename, 'w') as _:
    pass

for i in range(100):
    save(i)
```
|  netcdf4    |  h5netcdf 0.12.0  |  h5netcdf-refactor |
|-----|------|-----|
| 2.6 sec   | 21.2 sec | 3.6 sec |

shoyer (approved):

kmuehlbauer:
Let's merge #104 first and rebase here. It might make regression detection easier (not hoping there will be any).

PR #117, opened 2021-12-17
Avoid circular references to objects referencing h5py objects

TomAugspurger:
As discussed in https://github.com/h5py/h5py/issues/2019, using h5netcdf with file-like objects internally use threads can cause deadlocks with Python's GC collects objects with circular references.

This graph, generated with pympler, shows the circular objects created by `h5netcdf.File(...)`.

h5py-circular-ref

On `master`, this script will deadlock:

```python
import fsspec
import h5netcdf
import h5py


def main():
    url = "https://pangeo.blob.core.windows.net/pangeo-public/hurs_day_ACCESS-CM2_historical_r1i1p1f1_gn_1950.nc"
    for i in range(100):
        with fsspec.open(url).open() as o:
            h5netcdf.core.File(o, mode="r", decode_vlen_strings=True)

        print("--------------------------------------", i)


if __name__ == "__main__":
    main()
```

This PR changes all the circular references I could find to use `weakref`s. With this fix, that script seems to consistently run until completion.


TomAugspurger:
@shoyer or @kmuehlbauer are either of you able to approve the CI run when you get a chance? Things passed for me locally with Python 3.8, but I haven't tested on other versions yet.

kmuehlbauer:
@TomAugspurger Thanks for looking into this. It looks like the linter isn't happy yet. I'll have a look at this tomorrow morning (CET).

TomAugspurger:
Great thanks! No rush on this.

76160a2777b2b54fe407449ef3e884c3db3440d3 hopefully fixes the linting. The h5py2 builds seem to be failing to import 
 netCDF4 (https://github.com/h5netcdf/h5netcdf/runs/4565072755?check_suite_focus=true)

shoyer:
@TomAugspurger thanks for putting this together!

I don't think it's worth diving into the failing tests with h5py version 2. Those are almost certainly already failing on `main`.

shoyer (approved):

tacaswell:
I am _mildly_ interested in what broke between h5py2 and h5py3.....

shoyer:
The only thing that changed is that the conda-forge installation of our h5py version 2 setup is broken. My guess is we could fix this by pinning to older versions of netCDF4 and/or libnetcdf, so they are built against the same version of HDF5.

kmuehlbauer:
@TomAugspurger let's get this in and I'll take care of the h5py2 issue in a follow up PR. Thanks a bunch!

kmuehlbauer:
@TomAugspurger `h5netcdf` release `0.12.0` with your fixes has just been released and is now available on PyPI.

TomAugspurger:
Great, thanks!

PR #118, opened 2021-12-19
Pinned netCDF4

TomAugspurger:
This "fixes" the CI failure by pinning netCDF to a version that seems to be compatible with whatever version of h5py and libnetcdf that mamba picks. netCDF4==1.5.3 was released around the same time as h5py==2.10.0.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -55,7 +55,9 @@ jobs:
         python-version: ${{ matrix.python-version }}
     - name: Set up conda environment
       run: |
-        mamba install pip wheel pytest h5py==2.10.0 netCDF4
+        # pinned netCDF4==1.5.3 to avoid an issue when trying to import newer netCDF4 versions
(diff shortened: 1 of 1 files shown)


TomAugspurger:
@kmuehlbauer this is passing now. I'm not really sure what's going on between netCDF and libnetcdf, but hopefully pinning to an older netCDF is an acceptable workaround.

---

I have no idea what the release process if like for h5netcdf, but getting a release out with #117 would be very helpful for projects like pangeo-forge.

kmuehlbauer:
Thanks for setting this up, @TomAugspurger. I think we can cut a release in near time, next week.

kmuehlbauer (approved):

kmuehlbauer:
@TomAugspurger The release process is very simple. Just doing a github release will trigger the build and upload to PyPI.

I didn't had time at the weekend to give this more love, but I'll merge this tomorrow.

Before releasing the Changelog and version file needs changing. I can do this tomorrow, too. So we could do the release sometime tomorrow.

@shoyer Is there anything blocking a release tomorrow? Something I might have forgotten?

kmuehlbauer:
Thanks again @TomAugspurger!

PR #120, opened 2021-12-21
Fixed one more cycle

TomAugspurger:
Apologies for missing this in #117, but there was at least one more cycle in h5netcdf that my test case didn't pick up. When used via xarray, this seems to come up.

image

It's apparently created by a cycle in `_determine_current_dimension_size`, by the closure `find_dim`. I'm not sure exactly what cycle is, and so I'm struggling to write a test to detect it. But in the absence of a test, I'm running a more realistic workload against this branch now that'll hopefully detect any issues. I'll try to come up with a test in the meantime.

Apologies again for pushing for a release before really testing stuff out :/

### DIFF
--- a/h5netcdf/core.py
+++ b/h5netcdf/core.py
@@ -376,11 +376,6 @@ class Group(Mapping):
         if self.dimensions[dim_name] is not None:
             return max_size
 
-        def _find_dim(h5group, dim):
-            if dim not in h5group:
-                return _find_dim(h5group.parent, dim)
-            return h5group[dim]
-
         dim_variable = _find_dim(self._h5group, dim_name)
 
         if "REFERENCE_LIST" not in dim_variable.attrs:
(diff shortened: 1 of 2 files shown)


TomAugspurger:
Still seeing hangs on this branch :/ So perhaps there's another cycle, or this cycle is a red herring, or both, or something else entirely.

I'll keep exploring...

kmuehlbauer:
@TomAugspurger No worries. If I can be of any help, please let me know.

kmuehlbauer:
@TomAugspurger Would it be possible to create a minimal example? I'm also interested in how to use `pympler` to extract/find those cycles. It will be useful for future checks, too.

kmuehlbauer:
@TomAugspurger Do you still experience issues with dangling references? There have been several merges since this PR, you might need to rebase to current master.

TomAugspurger:
Apologies for leaving this. I haven't been able to reproduce the hanging again. I haven't done any jobs as large as I was, so most likely I thought I had a version with https://github.com/h5netcdf/h5netcdf/pull/117 but actually didn't.

For creating these references, I'd actually recommend using [`objgraph`](https://objgraph.readthedocs.io/en/stable/) rather than `muppy` now. I had to patch muppy because one of its functions was raising when trying to access the size of one of these objects.

Going to close this, and I'll reopen / investigate more if I am able to reproduce it again.

Issue #68, opened 2020-01-10
Group.name property returns full path instead of name


kmuehlbauer:
XRef:  https://github.com/pydata/xarray/issues/3680 for more detail.

Sorry, if that has been discussed before.

I see two points here:

- alignment with h5py which returns the full name of the object (full path from root), status quo
- alignment with netCDF4 which returns the name

Xarray tries to use `h5netcdf` in the same manner as `netCDF4`, which breaks due to this.

The questions is, where should this be fixed? Thanks for any ideas!

kmuehlbauer:
Time to find a solution to this issue. 

```python
import netCDF4
import h5netcdf.legacyapi as h5nc4
with netCDF4.Dataset("testx.nc", "w", format="NETCDF4") as root:
    test1 = root.createGroup("test1")
    test1.createGroup("test2")
    print(root['/test1/test2'].name)
with h5nc4.Dataset('testx.nc') as h5n:
    print(h5n["/test1/test2"].name)

test2
/test1/test2
```

So, at least for the legacy API we should try to conform to netCDF4. This will not help https://github.com/pydata/xarray/issues/3680, but I do see no way to resolve this in another way (besides breaking conformity with `h5py`). Do you have any preferences, @shoyer?

kmuehlbauer:
resolved in #126

PR #126, opened 2021-12-23
return group name instead of full path name for legacy API

kmuehlbauer:
- [x] Closes Issue #68 
- [x] Tests added

### DIFF
--- a/h5netcdf/core.py
+++ b/h5netcdf/core.py
@@ -289,7 +289,7 @@ class Group(Mapping):
     def __init__(self, parent, name):
         self._parent_ref = weakref.ref(parent)
         self._root_ref = weakref.ref(parent._root)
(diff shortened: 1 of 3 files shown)


kmuehlbauer:
It's a small fix to align legacy API with netcdf4-python. I'll merge in a couple of days.

h5netcdf/core.py
```diff
@@ -289,7 +289,7 @@ def _group_cls(self):
     def __init__(self, parent, name):
         self._parent_ref = weakref.ref(parent)
         self._root_ref = weakref.ref(parent._root)
-        self._h5path = _join_h5paths(parent.name, name)
+        self._h5path = _join_h5paths(parent._h5path, name)
```
kmuehlbauer, line 292: This always gets the full path for constructing the h5path since .name gives back group name only for legacyapi.

Issue #52, opened 2019-03-06
Humongous netCDF files with unlimited dimensions


dionhaefner:
I am the author of a pure-Python ocean model. After switching from netCDF4 to h5netcdf I noticed that output files were much larger.

This seems to be the case only when using unlimited dimensions, and the file does not grow by much when the dimension is resized later on, so I guess it is down to how much space is preallocated for unlimited dimensions.

Example script to reproduce the problem:

```python
import os
import tempfile

import h5netcdf
import netCDF4

import numpy as np


DIMENSIONS = {
    "x": 250,
    "y": 250,
    "z": 40,
    "time": None,
}

VAR_DIMS = ["time", "z", "y", "x"]


def create_h5netcdf_file(outfile):
    with h5netcdf.File(outfile, "w") as ncfile:
        for dim, size in DIMENSIONS.items():
            ncfile.dimensions[dim] = size
            ncfile.create_variable(dim, (dim,), "float64")

        ncfile.create_variable("test", VAR_DIMS, "float64")

        if DIMENSIONS["time"] is None:
            ncfile.resize_dimension("time", 1)

        ncfile.variables["time"][0] = 1
        ncfile.variables["test"][:] = np.random.rand(
            *(DIMENSIONS[dim] or 1 for dim in VAR_DIMS)
        )


def create_netcdf4_file(outfile):
    with netCDF4.Dataset(outfile, "w") as ncfile:
        for dim, size in DIMENSIONS.items():
            ncfile.createDimension(dim, size)
            ncfile.createVariable(dim, "f8", dim)

        ncfile.createVariable("test", "f8", VAR_DIMS)
        ncfile.variables["time"][0] = 1
        ncfile.variables["test"][:] = np.random.rand(
            *(DIMENSIONS[dim] or 1 for dim in VAR_DIMS)
        )


def get_filesize_mb(infile):
    return os.stat(infile).st_size // 1024 ** 2


if __name__ == "__main__":
    outfile = tempfile.NamedTemporaryFile(delete=False)
    outfile.close()

    try:
        create_h5netcdf_file(outfile.name)
        print("h5netcdf: %sMB" % get_filesize_mb(outfile.name))
    finally:
        os.remove(outfile.name)

    outfile = tempfile.NamedTemporaryFile(delete=False)
    outfile.close()

    try:
        create_netcdf4_file(outfile.name)
        print("netcdf4: %sMB" % get_filesize_mb(outfile.name))
    finally:
[...]

PR #127, opened 2021-12-28
implement new h5netcdf default chunksize

dionhaefner:
- [x] Closes Issue #52 
- [x] Tests added

### DIFF
--- a/h5netcdf/core.py
+++ b/h5netcdf/core.py
@@ -482,7 +482,7 @@ class Group(Mapping):
         return group._create_child_group(keys[-1])
 
     def _create_child_variable(
-        self, name, dimensions, dtype, data, fillvalue, chunks, **kwargs
+        self, name, dimensions, dtype, data, fillvalue, **kwargs
     ):
         if name in self:
             raise ValueError(
@@ -529,10 +529,6 @@ class Group(Mapping):
         if shape != maxshape:
             kwargs["maxshape"] = maxshape
 
-        # Determine default chunksize if not given
-        if chunks is None:
-            chunks = self._get_default_chunksize(dimensions, dtype)
-
         # Clear dummy HDF5 datasets with this name that were created for a
         # dimension scale without a corresponding variable.
         # Keep the references, to re-attach later
@@ -544,7 +540,7 @@ class Group(Mapping):
                 self._delete_dim_scale(name)
 
         self._h5group.create_dataset(
-            h5name, shape, dtype=dtype, data=data, fillvalue=fillvalue, chunks=chunks, **kwargs
+            h5name, shape, dtype=dtype, data=data, fillvalue=fillvalue, **kwargs
         )
 
         self._variables[h5name] = self._variable_cls(self, h5name, dimensions)
@@ -571,84 +567,19 @@ class Group(Mapping):
             variable.attrs._h5attrs["_FillValue"] = value
         return variable
 
-    def _get_default_chunksize(self, dimensions, dtype):
-        # This is ported from
-        # https://github.com/Unidata/netcdf-c/blob/a57101d4b7dcabf7d477c08eee9a5a126732f702/libhdf5/hdf5var.c#L104-L225
-        # (published under BSD-3-Clause)
-
-        DEFAULT_CHUNK_SIZE = 16_777_216
-        DEFAULT_1D_UNLIM_SIZE = 4096
-
-        ndims = len(dimensions)
-        chunks = [None] * ndims
-
-        num_values = 1
-        num_unlimited = 0
-
-        type_size = np.dtype(dtype).itemsize
-
-        # How many values in the variable (or one record, if there are unlimited dimensions).
-        for i, d in enumerate(dimensions):
-            dimsize = self.dimensions[d]
-            if dimsize is not None:
-                num_values *= dimsize
-            else:
-                num_unlimited += 1
-                chunks[i] = 1
-
-        # Special case to avoid 1D vars with unlim dim taking huge amount
-        # of space (DEFAULT_CHUNK_SIZE bytes). Instead we limit to about
-        # 4KB.
-        if ndims == num_unlimited == 1:
-            suggested_size = max(1, min(DEFAULT_1D_UNLIM_SIZE, DEFAULT_CHUNK_SIZE // type_size))
-            chunks[0] = suggested_size
-
-        elif ndims == num_unlimited > 1:
-            # all dims unlimited
-            suggested_size = int((DEFAULT_CHUNK_SIZE / type_size) ** (1 / ndims))
-            chunks = [suggested_size] * ndims
-
-        # Pick a chunk length for each dimension, if one has not already been picked above.
-        for i, d in enumerate(dimensions):
-            if chunks[i] is not None:
-                continue
-
-            dimsize = self.dimensions[d]
-            suggested_size = min(
-                (DEFAULT_CHUNK_SIZE / (num_values * type_size)) ** (1 / (ndims - num_unlimited))
-                * dimsize - 0.5,
-                dimsize
-            )
-            chunks[i] = int(suggested_size)
-
-        # Do we have any big data overhangs? They can be dangerous to
-        # babies, the elderly, or confused campers who have had too much
-        # beer.
-        for i, d in enumerate(dimensions):
-            dimsize = self.dimensions[d]
-            if dimsize is None:
-                continue
-
-            chunksize = chunks[i]
-            num_chunks = (dimsize + chunksize - 1) // chunksize
-            overhang = num_chunks * chunksize - dimsize
-            chunks[i] -= overhang // num_chunks
-
-        return tuple(chunks)
-
     def create_variable(
-        self, name, dimensions=(), dtype=None, data=None, fillvalue=None, chunks=None, **kwargs
+        self, name, dimensions=(), dtype=None, data=None, fillvalue=None, **kwargs
     ):
         if name.startswith("/"):
             return self._root.create_variable(
-                name[1:], dimensions, dtype, data, fillvalue, chunks, **kwargs
+                name[1:], dimensions, dtype, data, fillvalue, **kwargs
             )
         keys = name.split("/")
         group = self
         for k in keys[:-1]:
             group = group._require_child_group(k)
         return group._create_child_variable(
-            keys[-1], dimensions, dtype, data, fillvalue, chunks, **kwargs
+            keys[-1], dimensions, dtype, data, fillvalue, **kwargs
         )
 
     def _get_child(self, key):
@@ -1037,3 +968,83 @@ class File(Group):
             self.mode,
         )
         return "\n".join([header] + self._repr_body())
+
+
+def _get_default_chunksizes(dimsizes, dtype):
+    # This is ported from
+    # https://github.com/Unidata/netcdf-c/blob/a57101d4b7dcabf7d477c08eee9a5a126732f702/libhdf5/hdf5var.c#L104-L225
+    # (published under BSD-3-Clause)
+
+    DEFAULT_CHUNK_SIZE = 16_777_216
+    DEFAULT_1D_UNLIM_SIZE = 4096
+    H5_MAX_CHUNK_SIZE = 4_294_967_295
+
+    ndims = len(dimsizes)
+    chunks = [None] * ndims
+
+    num_values = 1
+    num_unlimited = 0
+
+    type_size = np.dtype(dtype).itemsize
+
+    # How many values in the variable (or one record, if there are unlimited dimensions).
+    for i, dimsize in enumerate(dimsizes):
+        if dimsize is not None:
+            num_values *= dimsize
+        else:
+            num_unlimited += 1
+            chunks[i] = 1
+
+    # Special case to avoid 1D vars with unlim dim taking huge amount
+    # of space (DEFAULT_CHUNK_SIZE bytes). Instead we limit to about
+    # 4KB.
+    if ndims == num_unlimited == 1:
+        suggested_size = max(
+            1, min(DEFAULT_1D_UNLIM_SIZE, DEFAULT_CHUNK_SIZE // type_size)
+        )
+        chunks[0] = suggested_size
+
+    elif ndims == num_unlimited > 1:
+        # all dims unlimited
+        suggested_size = int((DEFAULT_CHUNK_SIZE / type_size) ** (1 / ndims))
+        chunks = [suggested_size] * ndims
+
+    # Pick a chunk length for each dimension, if one has not already been picked above.
+    for i, dimsize in enumerate(dimsizes):
+        if chunks[i] is not None:
+            continue
+
+        suggested_size = min(
+            (DEFAULT_CHUNK_SIZE / (num_values * type_size))
+            ** (1 / (ndims - num_unlimited))
+            * dimsize
+            - 0.5,
+            dimsize,
+        )
+        chunks[i] = int(suggested_size)
+
+    # But did this result in a chunk that is too big?
+    def check_chunksizes(chunks):
+        total_chunksize = type_size
+        for c in chunks:
+            total_chunksize *= c
+
+        return total_chunksize <= H5_MAX_CHUNK_SIZE
+
+    while not check_chunksizes(chunks):
+        # Chunk is too big! Reduce each dimension by half and try again.
+        chunks = [max(1, c // 2) for c in chunks]
+
+    # Do we have any big data overhangs? They can be dangerous to
+    # babies, the elderly, or confused campers who have had too much
+    # beer.
+    for i, dimsize in enumerate(dimsizes):
+        if dimsize is None:
+            continue
+
+        chunksize = chunks[i]
+        num_chunks = (dimsize + chunksize - 1) // chunksize
+        overhang = num_chunks * chunksize - dimsize
+        chunks[i] -= overhang // num_chunks
+
+    return tuple(chunks)
--- a/h5netcdf/core.py
+++ b/h5netcdf/core.py
@@ -669,7 +669,8 @@ class Group(Mapping):
                 chunks = _get_default_chunksizes(maxshape, dtype)
             else:
                 raise ValueError(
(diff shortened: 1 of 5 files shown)


dionhaefner:
For the script from #52 this prints:

```bash
h5netcdf: 19MB
netcdf4: 19MB
```

as expected.

---

Some tests are failing but I wanted to get your feedback on some issues first: 

1. Should this change the default only for the legacy API or for the new API, too?
2. There is some logic [that ensures the number of bytes per chunk is below `NC_MAX_UINT`](https://github.com/Unidata/netcdf-c/blob/a57101d4b7dcabf7d477c08eee9a5a126732f702/libhdf5/hdf5var.c#L195-L207) (4_294_967_295 by default). I assumed we don't need that so I didn't port it. Do you agree?

kmuehlbauer:
@dionhaefner Nixe! Let's see how the tests run. Maybe also add a test which checks for correct chunk sizes after file creation? 

To your questions:

1. It should only be default for the legacy API.

2. Not sure. Is this a hard restriction of the netCDF4 standard? 

Another thing with respect to the ported code -> we might have to add the netcdf-c license text somewhere. @shoyer recommended to put the code in a separate file in such occasions. But let's wait for his opinion.

dionhaefner:
> 1. It should only be default for the legacy API.

That sounds confusing to me, but well, it's your library. In this case this should probably be documented somewhere? Right now the README makes it sound like old and new API are functionally equivalent.

> 2. Not sure. Is this a hard restriction of the netCDF4 standard?

Looks more like this is [a limitation of HDF5](https://support.hdfgroup.org/HDF5/doc/Advanced/Chunking/):

> The maximum number of elements in a chunk is 2^32-1 which is equal to 4,294,967,295.

If I try to create a bigger chunk, `h5py` throws an exception:

```python
>>> f.create_dataset("foo", shape=(int(5e9),), chunks=(5e9,))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/dion/.pyenv/versions/h5netcdf/lib/python3.8/site-packages/h5py/_hl/group.py", line 149, in create_dataset
    dsid = dataset.make_new_dset(group, shape, dtype, data, name, **kwds)
  File "/home/dion/.pyenv/versions/h5netcdf/lib/python3.8/site-packages/h5py/_hl/dataset.py", line 109, in make_new_dset
    dcpl = filters.fill_dcpl(
  File "/home/dion/.pyenv/versions/h5netcdf/lib/python3.8/site-packages/h5py/_hl/filters.py", line 256, in fill_dcpl
    plist.set_chunk(chunks)
  File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper
  File "h5py/_objects.pyx", line 55, in h5py._objects.with_phil.wrapper
  File "h5py/h5p.pyx", line 473, in h5py.h5p.PropDCID.set_chunk
ValueError: All chunk dimensions must be less than 2^32 (all chunk dimensions must be less than 2^32)
```

I'll patch in the code path for big chunks, then.

shoyer:
What does h5py use for chunksize heuristics by default?

Generally we try to make the new API more "h5py-like" and the legacy API more "netCDF4-like" but that isn't a hard rule, and it's nice to be able to achieve the same behavior with both APIs.

dionhaefner:
> What does h5py use for chunksize heuristics by default? Are there cases where users would prefer that?

The heuristic is here:

https://github.com/h5py/h5py/blob/1d569e69fb8d2b1abea196792bb1f8c948180534/h5py/_hl/filters.py#L338-L389

Basically it starts with the dimensions of the dataset (or a fixed 1024 for unlimited axes) and divides each iteratively by 2 until it is close enough to the target chunksize (between 8K and 1M).

IMO the biggest issue is that unlimited dimensions default to a chunk size >> 1. In any NetCDF file I've encountered, unlimited dimensions are something like time where you just extend by 1 at a time.

kmuehlbauer:
tl;dr chunking heuristics is quite complex

@dionhaefner Just to summarize: the major pain point with the new API is that the unlimited dimensions can get a default chunk size (using h5py heuristics) much larger than 1 on creation of the dataset. This, in turn, can result in exploding filesizes, as follows:

Array size
----------
(1, 40, 250, 250)

Chunks
------

netcdf4
-------
(1, 20, 125, 125) -> 2500000 bytes per chunk
1 * (2 chunks * 20) * (2 chunks * 125) * (2 chunks * 125) * 8 byte= 20000000 (complete netcdf4 size)

h5netcdf
--------
(64, 3, 16, 16) ->  393216 bytes per chunk
64 * (14 chunks * 3) * (16 chunks * 16) * (16 chunks * 16) * 8 byte= 1409286144 (complete h5netcdf size)

Citing from Andrew Colletes Book (Python and HDF5):

> The “auto-chunker” tries to keep chunks mostly “square” (in N dimensions) and within certain size limits. 

> **Do you even need to specify a chunk size?**
>     It’s best to restrict manual chunk sizing to cases when you know for sure your dataset will be accessed in a way that’s likely to be inefficient with either contiguous storage or an auto-guessed chunk shape. And like all optimizations, you should benchmark! 
> **Try to express the “natural” access pattern your dataset will have**
>     As in our example, if you are storing a bunch of images in a dataset and know that your application will be reading particular 64×64 “tiles,” you could use N×64×64 chunks (or N×128×128) along the image axes. 
> **Don’t make them too small**
>     Keep in mind that HDF5 has to use indexing data to keep track of things; if you use something pathological like a 1-byte chunk size, most of your disk space will be taken up by metadata. A good rule of thumb for most datasets is to keep chunks above 10KiB or so. 
> **Don’t make them too big**
>     The key thing to remember is that when you read any data in a chunk, the entire chunk is read. If you only use a subset of the data, the extra time spent reading from disk is wasted. Keep in mind that chunks bigger than 1 MiB by default will not participate in the fast, in-memory “chunk cache” and will instead be read from disk every time.

I'm not sure what netcdf-c/netcdf4 has in the loop in terms of caching, but hdf5/h5py has a restriction on [cache size](https://docs.h5py.org/en/stable/high/file.html?highlight=chunk%20cache#chunk-cache) which can only be adapted at file-open time.

In our example case that would mean that performance might be degraded, since the chunk size is well above the 1MiB limit (2.5 MiB). On the other hand the chunk size for time dimension is definitive too big (64(. We can't just set it to 1, since this would change the size of one chunk to only 1 * 3 * 16 * 16 * 8 byte = 6144 bytes which is below the recommended lower thrshold of 10KiB.

If the conformity with hdf5/h5py is not a hard rule, as @shoyer pointed out, I think we have several options here. 

- keep everything as is
- switch to netcdf4 heuristics for legacy API, keep new API as is
- switch to netcdf4 heuristics for both API's
- use netcdf4 heuristics for legacy API, reimplement hdf5 heuristics for new API, but special case unlimited dimensions
- other

For the script in #52 I get (using this PR) the following chunk sizes:

"test" variable
h5netcdf: (64, 3, 16, 16)
legacy: (1, 20, 125, 125)
netcdf4: (1, 20, 125, 125)

"time" variable
h5netcdf: (1024,)
legacy: (4096,)
netcdf4: (512,)

"z", "y", "x" variables
h5netcdf: None, None, None
legacy: (40,), (250,), (250,)
netcdf4: contiguous, contiguous, contiguous

@dionhaefner It seems that the coordinate variables are treated a bit different with the ported heuristics than original netcdf4.

dionhaefner:
🤔 I see. I'll have a look at the coordinate variables, perhaps just some integer / float division issue.

> In our example case that would mean that performance might be degraded, since the chunk size is well above the 1MiB limit (2.5 MiB).

I'd like to point out that the 1MB cache is not *per chunk*, but *for the whole dataset*:

> rdcc_nbytes sets the total size (measured in bytes) of the raw data chunk cache for each dataset. The default size is 1 MB. This should be set to the size of each chunk times the number of chunks that are likely to be needed in cache.

So whether your chunks are 300kiB or 2.5MiB doesn't really matter in any reasonable application since even in the best case you can only hold 3 chunks in the cache anyway, and unless you are reading these over and over again it doesn't matter. 

But that being said I'd like to add another possibility:

- Use the netCDF4 heuristic with an added upper limit of 1MiB like in the `h5py` heuristic for both new and legacy API (i.e., sort of a hybrid heuristic that respects `h5py` cache defaults)

kmuehlbauer:
> > In our example case that would mean that performance might be degraded, since the chunk size is well above the 1MiB limit (2.5 MiB).
> 
> I'd like to point out that the 1MB cache is not _per chunk_, but _for the whole dataset_:

Yes, sorry, you're completely right. I did not read careful enough. 

> But that being said I'd like to add another possibility:
> 
>     * Use the netCDF4 heuristic with an added upper limit of 1MiB like in the `h5py` heuristic for both new and legacy API (i.e., sort of a hybrid heuristic that respects `h5py` cache defaults)

That sounds reasonable. Would you take a stab at implementing it? You might enhance the current example and convert it to a test, too.

 For the time being I would also suggest to keep current behaviour, the netcdf-heuristics and your combined approach available codewise and decide which combination should be used after a final discussion together with @shoyer. 

We might have to think about some FutureWarning/DeprecationCycle anyway.

dionhaefner:
[Actually, netCDF4 increases cache size to 32MB by default](https://www.unidata.ucar.edu/software/netcdf/workshops/most-recent/nc4chunking/Cache.html):

> The cache size for data transfers in HDF5 is 1 MB, by default. This is small for many real applications, so netCDF-4 uses a default cache size of 32 MB.

Hence the target chunk size of 16MB in the netCDF4 heuristic. Are you sure you want to stick with the HDF5 default cache size of 1MB?

kmuehlbauer:
Looks like we've opened a can of worms. :grimacing: We might have to open files with `rdcc_nbytes=CACHESIZE`, to make use of it, right?

dionhaefner:
Yes that's what I had in mind.

I think 1MB sounds awfully small for modern hardware, and I'm sure the netCDF4 devs benchmarked this properly, so I'd probably increase it. But I understand if you want to stay compliant with HDF5, and I don't think it's going to make a huge difference in practice.

kmuehlbauer:
I do not sea a huge problem if we use this for the legacy API, as it should align with netcdf4 anyway. But I'm uncertain how to handle this for the new API. I'd wait for @shoyer's insight, which way this should be implemented in a future proof maybe parameterizable way.

dionhaefner:
I did some quick and dirty benchmarking with different options, and as it turns out, the `h5py` default settings actually perform best:

```bash
{'compression': None, 'chunks': True, 'rdcc_nbytes': 1048576}
write: 0.686s
size: 200MB
chunksize: (2, 5, 32, 32)
read: 0.104s
---
{'compression': None, 'chunks': True, 'rdcc_nbytes': 33554432}
write: 0.712s
size: 200MB
chunksize: (2, 5, 32, 32)
read: 0.117s
---
{'compression': None, 'chunks': (4, 20, 125, 125), 'rdcc_nbytes': 1048576}
write: 0.763s
size: 228MB
chunksize: (4, 20, 125, 125)
read: 0.202s
---
{'compression': None, 'chunks': (4, 20, 125, 125), 'rdcc_nbytes': 33554432}
write: 0.768s
size: 228MB
chunksize: (4, 20, 125, 125)
read: 0.120s
---
{'compression': 'gzip', 'chunks': True, 'rdcc_nbytes': 1048576}
write: 7.375s
size: 180MB
chunksize: (2, 5, 32, 32)
read: 0.920s
---
{'compression': 'gzip', 'chunks': True, 'rdcc_nbytes': 33554432}
write: 7.494s
size: 180MB
chunksize: (2, 5, 32, 32)
read: 0.977s
---
{'compression': 'gzip', 'chunks': (4, 20, 125, 125), 'rdcc_nbytes': 1048576}
write: 9.399s
size: 179MB
chunksize: (4, 20, 125, 125)
read: 1.100s
---
{'compression': 'gzip', 'chunks': (4, 20, 125, 125), 'rdcc_nbytes': 33554432}
write: 9.204s
size: 179MB
chunksize: (4, 20, 125, 125)
read: 1.035s
```

Now this is for the trivial case of a file with dimensions `(10, 40, 250, 250)` that is written and read in one go (no slicing involved), but still...

Given that we might just use the `h5py` heuristic, add a special case for unlimited dimensions and call it a day?

kmuehlbauer:
> Given that we could just use the `h5py` heuristic, add a special case for unlimited dimensions and call it a day? 

Yes, from the implementation work, this would be the easiest. 

To be as versatile as possible we could override the handling of the `chunks` kwarg in `create_dataset` like that:

```python
chunks = True # default, same as chunks = "h5py"
chunks = "h5py" # use default h5py/hdf5 heuristics
chunks = "netcdf4" # use netcdf4 heuristics
chunks = "h5netcdf" # use default h5py/hdf5 heuristics with special casing unlimited dimensions
```

If our aim is to be very much in line with netcdf-c/netcdf4-python we could set the default for legacy API to `chunks = "netcdf4"`.

That way we keep the current behaviour (no deprecations etc.) and enable the user to decide which heuristics to use. 

> Or even ask the h5py folks whether they could consider changing the default?

I think we should discuss your findings with the h5py people. Maybe they have already a suggestion or issue about that?

dionhaefner:
This is a pretty neat API!

I have opened h5py/h5py#2029, let's see what they say :)

dionhaefner:
[Looks like they are hesitant to merge this.](https://github.com/h5py/h5py/pull/2036). Should we go ahead and implement it in `h5netcdf`, with your proposed API?

kmuehlbauer:
@dionhaefner 

Yes, I think we can add this here. If this will eventually be used enough, this might get enough interest upstream. I completely understand h5py-devs, as use-cases are widespread and the number people requesting this particular feature low. 

```python
chunks = True # default, same as chunks = "h5py"
chunks = "h5py" # use default h5py/hdf5 heuristics
# chunks = "netcdf4" # use netcdf4 heuristics
chunks = "h5netcdf" # use default h5py/hdf5 heuristics with special casing unlimited dimensions
```

I would not implement the netcdf4 heuristics for now.

dionhaefner:
How about issuing a warning if `chunks=None` or `chunks=True` and unlimited dimensions are present? E.g.

```python
>>> f.create_dataset("foo", shape=(None, 1, 2, 3))
UserWarning: Using h5py's default chunking with unlimited dimensions can lead 
to larger file sizes and degraded performance. Consider using `chunks="h5netcdf"` 
or explicit chunk sizes. To silence this warning, pass `chunks="h5py"`.
```

kmuehlbauer:
Could we equip this warning with both of the chunk sizes? The user might ideally see, which one to choose for the setup.

I'm still a bit hesitant, since this will clutter the logs downstream and code changes are needed everywhere. As `h5netcdf` is supposed to reflect netcdf4-python behavior anyway we could instead state that `h5netcdf` will be the default in later versions (1.0 ?).

dionhaefner:
I actually think changing the defaults (after a depreciation cycle) is the best way forward, so if you're open to that I'm happy :)

kmuehlbauer:
Yes, I do not see an easier way to move on, if we want to make people aware of that. Having that warning there forever is a no-go.

dionhaefner:
Example warning:

> FutureWarning: Using h5py's default chunking with unlimited dimensions can lead to increased file sizes and degraded performance (using chunks: (2, 3, 3, 256)). Consider passing ``chunks="h5netcdf"`` (would give chunks: (10, 10, 10, 1); default in h5netcdf >= 1.0), or set chunk sizes explicitly. To silence this warning, pass ``chunks="h5py"``.

Is there anything else you would like covered by tests regarding chunking or unlimited dimensions?

kmuehlbauer:
@dionhaefner I appreciate all your efforts. Rethinking the whole issue there might be a solution to our problem on the `h5py`-side (https://github.com/h5py/h5py/pull/2036#issuecomment-1016137566). I'm hoping that this can lead us to prevent the reimplementation of the autochunker code here.

dionhaefner:
IMO this would be a worse solution, since it wouldn't work in cases where you actually *want* unlimited chunks > 1. The heuristic should work for all of these shapes:

```python
shape = (None,)  # want chunks (1024,)
shape = (None, 1, 1)  # want chunks (1024, 1, 1)
shape = (None, None)  # want chunks (32, 32)
shape = (None, 128, 128)  # want chunks (1, 128, 128)
```

The proposed h5netcdf heuristic handles all of these, and more.

Anyhow, my bandwidth to work on this is exhausted. I'll unsubscribe for now, feel free to ping me if you want to take this up again or need my input.

kmuehlbauer:
That's why I was wondering if that would be feasible at all. Obviously it's not. 

@dionhaefner There are two things to do to get this in finally. Adding the h5py license into licences-folder (partly reimplementation, but would be good anyway) and fixing the linter error. I can do this, if your schedule doesn't allow it. Please let me know. Thanks.

dionhaefner:
That should do it :)

kmuehlbauer (approved):
@dionhaefner Welcome as a first time contributor and thanks for this enhancement.

h5netcdf/core.py
```diff
@@ -615,7 +615,7 @@ def create_group(self, name):
         return group._create_child_group(keys[-1])
 
     def _create_child_variable(
-        self, name, dimensions, dtype, data, fillvalue, **kwargs
+        self, name, dimensions, dtype, data, fillvalue, chunks, **kwargs
```
mahdiansyah1, line 618: import h5netcdf import numpy as np with h5netcdf.File('mydata.nc', 'w') as f: # set dimensions with a dictionary f.dimensions = {'x': 5} # and update them with a dict-like interface # f.dimensions['x'] = 5 # f.dimensions.update({'x': 5}) v = f.create_variable('hello', ('x',), float) v[:] = np.ones(5) # you don't need to create groups first # you also don't need to create dimensions first if you supply data # with the new variable v = f.create_variable('/grouped/data', ('y',), data=np.arange(10)) # access and modify attributes with a dict-like interface v.attrs['foo'] = 'bar' # you can access variables and groups directly using a hierarchical # keys like h5py print(f['/grouped/data']) # add an unlimited dimension f.dimensions['z'] = None # explicitly resize a dimension and all variables using it f.resize_dimension('z', 3)

mahdiansyah1 (approved):

Issue #128, opened 2022-01-09
Files written by h5netcdf cannot be edited by netcdf4-python


hmaarrfk:
#### What happened:

Files written by h5netcdf cannot be edited by netcdf4-python. It would be good if they could be. I believe the following patch is needed. Would you like me to make the Pull Request?

#### MCVE Code Sample

```python
import xarray as xr
import numpy as np
dataset = xr.DataArray(
    data=np.zeros((3, 3)),
    name="my_data"
)

engine = "h5netcdf"
# engine = "netcdf4"
dataset.to_netcdf("test.nc", engine=engine, format="NETCDF4")

import netCDF4
nc = netCDF4.Dataset("test.nc", mode="a")
```
```python
---------------------------------------------------------------------------
OSError                                   Traceback (most recent call last)
/tmp/ipykernel_367089/2679834096.py in <module>
      1 import netCDF4
----> 2 nc = netCDF4.Dataset("test.nc", mode="a")

src/netCDF4/_netCDF4.pyx in netCDF4._netCDF4.Dataset.__init__()

src/netCDF4/_netCDF4.pyx in netCDF4._netCDF4._ensure_nc_success()

OSError: [Errno -103] NetCDF: Can't write file: b'test.nc'
```

#### Expected Output
That it works.

#### Version

<details><summary>Output of <tt>print(h5py.version.info, f"\nh5netcdf {h5netcdf.__version__}")</tt></summary>

Summary of the h5py configuration
---------------------------------

h5py    3.6.0
HDF5    1.12.1
Python  3.9.9 | packaged by conda-forge | (main, Dec 20 2021, 02:41:03) 
[GCC 9.4.0]
sys.platform    linux
sys.maxsize     9223372036854775807
numpy   1.19.5
cython (built with) 0.29.24
numpy (built against) 1.19.5
HDF5 (built against) 1.12.1
 
h5netcdf 0.12.0

</details>

<details><summary>Suggested patch</summary>

```python
diff --git a/h5netcdf/core.py b/h5netcdf/core.py
index b68e3f5..5994928 100644
--- a/h5netcdf/core.py
+++ b/h5netcdf/core.py
@@ -462,7 +462,7 @@ class Group(Mapping):
     def _create_child_group(self, name):
         if name in self:
             raise ValueError("unable to create group %r (name already exists)" % name)
-        self._h5group.create_group(name)
+        self._h5group.create_group(name, track_order=True)
[...]

PR #131, opened 2022-01-11
release v0.13.0

kmuehlbauer:
Release 0.13.0 fixes bug #128 and improves performance #104, #112.


kmuehlbauer:
@shoyer I've set the release date to 12th of January (tomorrow). So if there is anything I've missed, we can add it here, before release.

Issue #134, opened 2022-01-12
Implement Dimension-class


kmuehlbauer:
#### Is your feature request related to a problem? Please describe.

Currently handling of dimensions in `h5netcdf` is cluttered over different `ChainMap`, which is sometimes hard to follow. That makes it hard to implement features (eg. `isunlimited()`) or just to nicely retrieve information from a single object.

#### Describe the solution you'd like

A `Dimension`-class which wraps around the underlying hdf5 dataset and which comprises all need functionality in connection with a single `ChainMap` containing a `Dimensions`-object for each group. With this `Dimension`-object (which might inherit from the same BaseClass as `Variable`) the extraction of all dimension information is done from a single object. All dimensions are accessible via the single `ChainMap` also from sub-groups.

#### Describe alternatives you've considered
I did not consider other approaches yet.

#### Additional context
With a dedicated `Dimension`-class the handling of dimensions internally is very much simplified. Additionally some implementations (eg. `phony_dim`) can be simplified too. Current code which handles dimensions can be moved to the new class, making all other classes cleaner.

I have a WIP implementation of the outlined concept ready and will link it shortly.

kmuehlbauer:
resolved with #135

Issue #36, opened 2017-10-18
legacy api: add isunlimited method


jhamman:
For compatability with netCDF4-Python, it would be nice to have the `isunlimited` on variable objects:

reference: http://unidata.github.io/netcdf4-python/#netCDF4.Dimension.isunlimited
xref: https://github.com/pydata/xarray/pull/1637

shoyer:
We don’t have notion of Dimension objects (yet) in h5netcdf. We just use integers, or None for unlimited.

shoyer:
I suppose this would indeed make sense for the legacy API.

kmuehlbauer:
> We don’t have notion of Dimension objects (yet) in h5netcdf. We just use integers, or None for unlimited.

A implementation of a Dimension-object is suggested in #134 and a WIP-PR is available #135 too.

kmuehlbauer:
resolved by #135

PR #135, opened 2022-01-12
Implement Dimension class

kmuehlbauer:
- [x] Closes Issue #134
- [x] Tests added

### DIFF
--- a/h5netcdf/core.py
+++ b/h5netcdf/core.py
@@ -3,7 +3,7 @@
 import os.path
 import warnings
 import weakref
-from collections import ChainMap, OrderedDict, defaultdict
+from collections import ChainMap, Counter, OrderedDict, defaultdict
 from collections.abc import Mapping
 
 import h5py
@@ -11,7 +11,7 @@ import numpy as np
 from packaging import version
 
 from .attrs import Attributes
-from .dimensions import Dimensions
+from .dimensions import Dimension, Dimensions
 from .utils import Frozen
 
 try:
@@ -35,10 +35,6 @@ _NC_PROPERTIES = "version=2,h5netcdf=%s,hdf5=%s,h5py=%s" % (
 )
 
 NOT_A_VARIABLE = b"This is a netCDF dimension but not a netCDF variable."
-
-
-def _reverse_dict(dict_):
-    return dict(zip(dict_.values(), dict_.keys()))
 
 
 def _join_h5paths(parent_path, child_path):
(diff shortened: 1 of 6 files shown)


kmuehlbauer:
@shoyer Please invite also other interested parties.

kmuehlbauer:
@mangecoeur - This adds dimension-objects #5
@jhamman - This adds the requested `isunlimited()`-method #36
@dionhaefner - This fixes erronous creation of dimensions - #138

You might have a look at the implementation, if there's interest. 

I'll let this sit for another couple of days before merge.

h5netcdf/core.py
```diff
@@ -646,16 +588,35 @@ def _create_child_variable(
...
+        # maybe create new dimensions depending on data
+        if data is not None:
+            for d, s in zip(dimensions, data.shape):
+                # create new dimensions only ever if
+                #  - they are not known via parent-groups
+                #  - they are given in dimensions
+                #  - it's not a coordinate variable, they will get special handling later
+                if d not in self._all_dimensions and d in dimensions and d is not name:
+                    # calls _create_dimension
+                    self.dimensions[d] = s
```
dionhaefner, line 600: This will still create invalid state if an exception is raised later, right? Maybe it would be worth wrapping this in `try ... except` or so?

kmuehlbauer: True, if a dimension is created here and something breaks afterwards, this would be the consequence.

I've used your example as test. At least this case is fixed. But you are right, we should take care of likely breakages.

Do you can think of something  which breaks with this implementation? Refactoring this function is still on my personal todo list.

dionhaefner: For example users wanting to use `chunks="h5netcdf"`, but not knowing whether it's available:

```python
try:
   f.create_variable(..., chunks="h5netcdf")
except:
   # fall back to h5py chunking
   f.create_variable(..., chunks=True)

# oops, file is broken
```

Or more general, pretty much every internal error in `h5py`'s `create_dataset`.

Seem like a somewhat simple fix would be this:

```python
orig_dimensions = self.dimensions.copy()

try:
    ...
except:
    # roll back changes
    self.dimensions = orig_dimensions
```

shoyer: 👍  if at all possible, it is much more user friendly to make "atomic" functions.

kmuehlbauer: @dionhaefner @shoyer I'd suggest to tackle that problem in a follow up PR. This one here is already quite huge. 

Issue #138 exist and we can coordinate there.

PR #137, opened 2022-01-15
MNT: move CI to micromamba, print test durations

kmuehlbauer:


kmuehlbauer:
I'm using micromamba in another project of mine already quite some time. It's blazing fast and creates the h5netcdf CI environments in roughly 30 seconds. The overall GH Actions runtime is reduced from approx 200 seconds to around 60 seconds. This really safes resources. :earth_africa: :earth_americas: :earth_asia:

Issue #147, opened 2022-01-27
Chunksize FutureWarning cannot be silenced without affecting "chunks"


kmuehlbauer:
#### What happened:

While fixing h5netcdf-changes over in xarray https://github.com/pydata/xarray/pull/6200 I found that those FutureWarnings cannot be silenced keeping current behavior.

- `chunks=None` -> `chunks=None` with `FutureWarning` 
- `chunks="h5py"` -> `chunks=True`
- `chunks="h5netcdf"` -> `chunks=_get_default_chunksizes`

#### What you expected to happen:

Silencing the `FutureWarning` keeping current behavior should be possible.

@dionhaefner Could you take a look here?

dionhaefner:
Actually I'm not quite sure whether we want to do `chunks="h5py"` -> `chunks=None`. The way I see it the proper migration is this:

```python
f.dimensions = {"x": 1, "time": None}

# OLD

# uses h5py chunking and throws warning
f.create_dataset("foo", ("x", "time"))  # could also be chunks=True

# no unlimited dims -> no warning
f.create_dataset("foo2", ("x",))  # could also be chunks=True

# NEW

# keep using h5py chunks, no warning
f.create_dataset("foo", ("x", "time"), chunks="h5py")

# no unlimited dims -> no warning (unchanged)
f.create_dataset("foo2", ("x",))
```

The distinction between `chunks="h5py"` and `chunks="h5netcdf"` is only relevant if unlimited dimensions are present. In this case, chunking is mandatory anyway. So the question boils down to: what do we want this to do?

```python
# no unlimited dimensions present, so this would be unchunked by default
f.create_dataset("foo2", ("x",), chunks="h5py")
```

I think the path of least confusion is to treat this like `chunks=True`. Since something is passed to the `chunks` keyword I would interpret this as "the user wants chunking". But if you disagree I'm fine with changing it to h5py's `chunks=None`.

kmuehlbauer:
@dionhaefner I appreciate this comprehensive representation of the different possibilities. I mostly agree and I think your outlined solution might be the way to go. But do you have a solution at hand for those users, who do not want to use the new default chunking after we switch to it?

dionhaefner:
That's essentially my last code snippet, which boils down to

1. If your variable has unlimited dimensions, use `chunks=h5py`.
2. If not, just keep using what you are using now.

This recipe should work in all cases to keep the current behavior.

[...]

PR #148, opened 2022-01-27
MNT: add "None" to chunks-kwarg to silence FutureWarning

kmuehlbauer:
- [x] Closes Issue #147
- [ ] Tests added

### DIFF
--- a/h5netcdf/core.py
+++ b/h5netcdf/core.py
@@ -621,11 +621,18 @@ class Group(Mapping):
         if shape != maxshape:
             kwargs["maxshape"] = maxshape
 
+        # keep track of chunks
+        use_chunks = chunks
+        if chunks is None:
+            use_chunks = None
+
         if isinstance(chunks, str):
             if chunks == "h5py":
-                chunks = True
+                use_chunks = True
             elif chunks == "h5netcdf":
-                chunks = _get_default_chunksizes(maxshape, dtype)
+                use_chunks = _get_default_chunksizes(maxshape, dtype)
+            elif chunks == "None":
+                use_chunks = None
             else:
                 raise ValueError(
                     "got unrecognized string value %s for chunks argument" % chunks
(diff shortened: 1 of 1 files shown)


h5netcdf/core.py
```diff
@@ -621,11 +621,18 @@ def _create_child_variable(
         if shape != maxshape:
             kwargs["maxshape"] = maxshape
 
+        # keep track of chunks
```
kmuehlbauer, line 624: This does not really help us. We can't overload the `chunks` to represent all possible situations:

- use h5py chunking, chunks=None, -> chunk only if maxshape is set
- use h5py chunking, chunks=True, -> chunk in any case
- use h5netcdf chunking, chunks=None, -> chunk only if maxshape is set
- use h5netcdf chunking, chunks=True, -> chunk in any case

Do you have a suggestion @dionhaefner?

kmuehlbauer: If we would add YAKWA `chunk_heuristics`, we could leave `chunks` as is and use `chunk_heuristics="h5netcdf"` to use h5netcdf-chunking and `chunk_heuristics="h5py"` for h5py-chunking. 

I do not see any other meaningful solution atm, so any input would be appreciated.

kmuehlbauer:
This will be a hell of a mess downstream to handle silencing that Futurewarning in a consistent way, as well for those who want to keep current behavour. It could be 10's to 100's and more of lines of code which needs fixing. 

I'm therefore inclined to keep the current behavior (h5py heuristics) instead of the planned switch at version 1.0 and make the new heuristics available via a global setting (file-level kwarg, config) and via a local setting (create_variable - level kwarg). That new kwarg should be advertised in the documentation.

The new kwarg could be `chunk_heuristic` or something along the lines.

Switching this behavior can be made anytime later, if we are really sure of the effects downstream. I'm sorry, but this seems to be the only way forward. 

@shoyer What is your suggestion here?

kmuehlbauer:
@dionhaefner If we decide to go the way outlined in https://github.com/h5netcdf/h5netcdf/issues/147#issuecomment-1025133024, I can change this PR accordingly.

kmuehlbauer:
closing in favour of #150

Issue #116, opened 2021-11-02
Vlen strings in attributes are not decoded after applying NCO


dionhaefner:
Example script:

```python
import h5netcdf
import netCDF4
import subprocess

with h5netcdf.File("test.nc", "w", decode_vlen_strings=True) as f:
    f.attrs["foo"] = "bar"

subprocess.run(["ncks", "test.nc", "-o", "test2.nc", "--overwrite"])

with h5netcdf.File("test.nc", "r", decode_vlen_strings=True) as f:
    print("attrs original file:")
    print(f.attrs)

with h5netcdf.File("test2.nc", "r", decode_vlen_strings=True) as f:
    print("\nattrs changed file:")
    print(f.attrs)

with netCDF4.Dataset("test2.nc", "r") as f:
    print("\nattrs changed file with netCDF4:")
    for attr in f.ncattrs():
        print(f"{attr}: {getattr(f, attr)!r}")

```

Output:

```bash
$ python h5netcdf_bug.py
attrs original file:
<class 'h5netcdf.attrs.Attributes'>
foo: 'bar'

attrs changed file:
<class 'h5netcdf.attrs.Attributes'>
foo: array(['bar'], dtype=object)
history: b'Tue Nov  2 11:57:35 2021: ncks test.nc -o test2.nc --overwrite'
NCO: b'netCDF Operators version 5.0.3 (Homepage = http://nco.sf.net, Code = http://github.com/nco/nco)'

attrs changed file with netCDF4:
foo: 'bar'
history: 'Tue Nov  2 11:57:35 2021: ncks test.nc -o test2.nc --overwrite'
NCO: 'netCDF Operators version 5.0.3 (Homepage = http://nco.sf.net, Code = http://github.com/nco/nco)'
```

In the changed file, `h5netcdf` doesn't decode the `foo` attribute anymore (`array(['bar'], dtype=object)`). NetCDF4 decodes everything.

NetCDF files: [ncfiles.zip](https://github.com/h5netcdf/h5netcdf/files/7460017/ncfiles.zip)

This is with most recent h5netcdf and h5py.

kmuehlbauer:
There is a difference in the underlying hdf-representation between those two files. The first has `SCALAR` DATASPACE whereas the second has `SIMPLE` (array). That's why an array is returned.

test.nc
```
ATTRIBUTE "foo" {
      DATATYPE  H5T_STRING {
         STRSIZE H5T_VARIABLE;
[...]

PR #151, opened 2022-02-06
fix array attributes and attributes decoding

kmuehlbauer:
- items are extracted from 0-dim and single element 1-dim attributes
- multi-element 1-dim attributes are returned as list (like netcdf4-python)
- bytes/string attributes are returned as strings decoded from their respective encoding (ascii/utf-8)

- [x] Closes Issue #116
- [x] Tests added

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -18,6 +18,10 @@ Development Version:
   #136). By `Mark Harfouche <h55ps://github.com/hmaarrfk>`_.
 - Implement Dimension-class.
   By `Kai Mühlbauer <https://github.com/kmuehlbauer>`_.
+- Return items from 0-dim and one-element 1-dim array attributes. Return multi-element
+  attributes as lists. Return string attributes as Python strings decoded from their respective
+  encoding (`utf-8`, `ascii`).
+  By `Kai Mühlbauer <https://github.com/kmuehlbauer>`_.
 
 Version 0.13.0 (January 12, 2022):
 
--- a/h5netcdf/attrs.py
+++ b/h5netcdf/attrs.py
@@ -26,12 +26,41 @@ class Attributes(MutableMapping):
 
         if key in _HIDDEN_ATTRS:
             raise KeyError(key)
+
+        # get original attribute via h5py low level api
+        # see https://github.com/h5py/h5py/issues/2045
+        attr = self._h5attrs.get_id(key)
+
         # see https://github.com/h5netcdf/h5netcdf/issues/94 for details
         if isinstance(self._h5attrs[key], h5py.Empty):
             string_info = h5py.check_string_dtype(self._h5attrs[key].dtype)
             if string_info and string_info.length == 1:
                 return b""
-        return self._h5attrs[key]
+
+        # transform to 1d array in any case for easy iteration
+        # see https://github.com/h5netcdf/h5netcdf/issues/116
+        output = np.atleast_1d(self._h5attrs[key])
+
+        # string decoding subtleties
+        # vlen strings are already decoded -> only decode fixed length strings
+        # see https://github.com/h5netcdf/h5netcdf/issues/116
+        # netcdf4-python returns string arrays as lists, we do as well
+        string_info = h5py.check_string_dtype(attr.dtype)
+        if string_info is not None:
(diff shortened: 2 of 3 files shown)


kmuehlbauer:
@dionhaefner Please have a look here for resolving #116. Tests need polishing and we need to decide what should only be attributed to the legacyapi (at least partly).

dionhaefner:
This only fixes the reading issue, right? So this will still produce NetCDF files with `SIMPLE` space for attributes?

dionhaefner:
As for the legacy API question, I think you will have to make up your mind what its the main use case is supposed to be. Is it:

1. A quick way to try out `h5netcdf` without refactoring to see if it makes sense for an application that currently uses netcdf4-python; or
2. A "compatibility mode" that stays as close to netcdf4-python as possible so users can just change the import and forget about it, with the only benefit of not requiring the NetCDF4 C library anymore.

In my mind the prime goal was (1), basically to supply an easy way to port existing code while cutting ties with NetCDF4-C baggage and suboptimal design choices. In this case the legacy API should have the same defaults as the new API. If the goal is actually (2) then it should stay compliant with NetCDF4.

kmuehlbauer:
> This only fixes the reading issue, right? So this will still produce NetCDF files with `SIMPLE` space for attributes?

Yes, reading only, for now. But non-array attributes are created with `SCALAR` space. Only array attributes are created with `SIMPLE`.

kmuehlbauer:
> In my mind the prime goal was (1), basically to supply an easy way to port existing code while cutting ties with NetCDF4-C baggage and suboptimal design choices. In this case the legacy API should have the same defaults as the new API. If the goal is actually (2) then it should stay compliant with NetCDF4 whenever possible.

@shoyer wrote in the README.rst

> The legacy API is designed for compatibility with [netCDF4-python](https://github.com/Unidata/netcdf4-python).

 There isn't much of a choice, if we want to stick with it. I see the benefit of having a legacyapi with same defaults as new API when working with existing code. We could introduce a switch to decide between `defaults` and `netcdf4` (only for `legacyapi` or for new api too). 
 
This is something which needs another set of eyes, @shoyer what is your opinion on that?

shoyer:
I agree that netCDF4 made some sub-optimal design choices in terms of how it uses HDF5, but I would hesitate to try to fix those in h5netcdf. The risk of creating files that are difficult or impossible to read with the netCDF-C library is very high.

So I guess my view is closer to (2), both the legacy API and the new API should be a drop-in replacements for netCDF4. The only case where I would consider using different design choices is if the user explicitly writes `invalid_netcdf=True`.

kmuehlbauer:
Thanks @shoyer for the input. I think it's a good idea to move the general discussion into a new issue to not clutter the PR.

For this PR (and #116) this means `h5netcdf` should read attributes properly (extracting single elements from attribute arrays). That would indeed solve #116. But it would leave us in the writing case with `netcdf4` files with attributes which are of SCALAR dataspace (which is non-standard according to [this comment](https://github.com/Unidata/netcdf-c/issues/127#issuecomment-292625916) from @DennisHeimbigner). In consequence we should try to write attributes with SIMPLE dataspace, instead.

dionhaefner:
Yes, and sounds like we agree that we should keep things consistent between old and new API? So this PR is fine as it is.

kmuehlbauer:
@dionhaefner There is another subtle issue here, see https://github.com/h5py/h5py/issues/2045.  That's a special case of a zero-dim fixed-size attribute, which can be written by `h5netcdf` and read by `netcdf4`. Reading / decoding isn't possible on `h5py`-side.

I'll try to add some tests, which xfail in these cases.

kmuehlbauer:
So, I'll have to fix some issues regarding `h5py2`/`h5py3` and with bytes string / string.

kmuehlbauer:
> So, I'll have to fix some issues regarding `h5py2`/`h5py3`

Still pending ...

kmuehlbauer:
That was no fun to get this running :-( @dionhaefner Would you mind having a look here again, especially at the tests and that your issue (#116) is really solved now.

With the current implementation everything is backwards compatible as far as h5netcdf tests are concerned. There might still be some glitches left, but they are either not that obvious or are rare edge cases.

PR #162, opened 2022-03-21
New way for testing with h5pyd

ajelenak:
Testing with h5pyd was not easy in the past due to dependency on a remote HSDS service. Since HSDS has added recently support for running locally, the changes here are enabling setup, startup, and shutdown of such HSDS service.

Small fixes to a few tests are included here also. They either enable use of h5pyd or avoid using the netcdf4 package for remote (HSDS) .nc files.

### DIFF
--- a/h5netcdf/tests/conftest.py
+++ b/h5netcdf/tests/conftest.py
@@ -5,17 +5,44 @@ from pathlib import Path
 from shutil import rmtree
 import pytest
 try:
+    from h5pyd._apps.hstouch import main as hstouch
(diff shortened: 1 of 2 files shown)


kmuehlbauer:
@ajelenak Great! Thanks for creating this PR. I can rebase or you can do if you prefer. Just let me know.

kmuehlbauer:
@ajelenak Currently the `hsds`/`h5pyd`-based tests are SKIPPED because both packages are not installed yet. But it works locally on my machine (I get the error you mentioned in the other issue).

Issue #182, opened 2022-09-18
very slow partial reading when saved with index shift


master-nemo:
#### What happened:
When saved two group of data with shifted index like in code fragment `***1`, partial reading will be very slow in future. Same index like `***2` - works normal. 
Note: displaying shape in sample is just to shorten the output. A normal reading to variable behaves similarly. 

upd2: fixed sample and notes. 

#### What you expected to happen:
fast reading for all cases

#### MCVE Code Sample

```python
 
# ------------------------------ saving part ----------------------------- #
import numpy as np
import h5netcdf.legacyapi as netCDF4
ofname=r'_o_.nc'
h,w=720,1440
reshap3=[w,h,1]

def calc1(*args): return np.random.random((h,w))
def calc_data_group2(*args): return np.random.random((h,w)),np.random.random((h,w))

import h5netcdf.legacyapi as netCDF4
with netCDF4.Dataset(ofname,'w',phony_dims='sort') as ds:
    
    x = ds.createDimension('x',w) 
    y = ds.createDimension('y',h) 
    q = ds.createDimension('q',None) 
    
    vx = ds.createVariable('x', 'f4', ('x',)) 
    vy = ds.createVariable('y', 'f4', ('y',)) 
    v_q = ds.createVariable('q', 'f8', ('q',)) 
    
    
    v_data1 = ds.createVariable('data', float, ('x','y','q') )

    v_U = ds.createVariable('U', float, ('x','y','q') )
    v_V = ds.createVariable('V', float, ('x','y','q') )
    
    q=0 
    prev=None
    for dat in range(60):   # trying for 300+,100,30 fragments of data. even for 30 reading is noticeble slow
        d1=calc1(dat)   #something like this with my data
        v_data1[:,:,q] = d1.reshape(reshap3)    #saving 1st vaiable 
            
        if q-2>=0: 
            U,V = calc_data_group2(prev,d1) #something like this with my data
            
            #                              saving second group of data
            # ***1
            v_U[:,:,q-2] = U[:,:].T.reshape(reshap3)  # saved at index q-2
            v_V[:,:,q-2] = V[:,:].T.reshape(reshap3)  # (!) very slow reading when saved like here
            # ***2
            # v_U[:,:,q] = U[:,:].T.reshape(reshap3)  # saved at index q
            # v_V[:,:,q] = V[:,:].T.reshape(reshap3)  # normal fast readig when saving at index q
        prev=d1
        q+=1    #main index of out data


# %%
[...]

PR #183, opened 2022-09-26
FIX: use default netcdf fillvalues, pad returned array if needed

kmuehlbauer:
- [x] Closes Issue #182 
- [x] Tests added


kmuehlbauer:
@master-nemo It would be great, if you could test your use-case against this PR. Thanks.

master-nemo:
> @master-nemo It would be great, if you could test your use-case against this PR. Thanks.

yes! it works for my old case! the reading is much faster now!
thanks!

Issue #143, opened 2022-01-21
Tracking issue for upstream order_tracking bug


hmaarrfk:
xref: https://github.com/h5py/h5py/issues/1385

It does somewhat seem like it is a bug in hdf5. I wonder if we can recreate it in netcdf4 and get them involved too!

kmuehlbauer:
I've come up with a workaround which also has the nice advantage to be a bit more performant when creating millions of attributes ;-)

https://github.com/h5py/h5py/pull/2042

FYI: @paugier

kmuehlbauer:
I've tested that patch a bit. So the `track_order` problem is avoided then. 

But a major pain point is, that every change of an attribute would delete and recreate it. With `track_order=True` this would move the changed attribute to the back of the list.

I think we need to fix this either in `h5py` (allowing to `modify` existing attributes with same dtype) or in `h5netcdf`. It seems, that netCDF4 keeps this order regardless of an attribute has changed value and/or type.

hmaarrfk:
Does netCDF4 suffer from the same issue? or no because they don't rename variable like h5py does?

kmuehlbauer:
Yes, netCDF4 suffers from the same issue. At least first tests suggest this.

kmuehlbauer:
```python
import netCDF4
with netCDF4.Dataset("test.nc", mode="w") as nc:
    g = nc.createGroup("grp0")
    g.test12 = 12
    g.test0 = 0
    g.test1 = 1
    g.test2 = 2
    g.test3 = 3
    g.test4 = 4
    g.test5 = 5
    g.test6 = 6
    g.test7 = 7
    g.test8 = 8
    g.test9 = 9
    g.test10 = 10
    g.test11 = 11

with netCDF4.Dataset("test.nc", mode="r") as nc:
    print(nc["grp0"])

with netCDF4.Dataset("test.nc", mode="a") as nc:
    g = nc["grp0"]
    g.test8 = 4
    g.test9 = "3"
    g.test1 = 42.
    
with netCDF4.Dataset("test.nc", mode="r") as nc:
    print(nc["grp0"])
```

```python
<class 'netCDF4._netCDF4.Group'>
group /grp0:
    test12: 12
    test0: 0
    test1: 1
    test2: 2
    test3: 3
    test4: 4
    test5: 5
    test6: 6
    test7: 7
    test8: 8
    test9: 9
    test10: 10
    test11: 11
    dimensions(sizes): 
[...]

PR #186, opened 2022-11-19
Make track_order True as default if h5py is 3.7.0 or greater

hmaarrfk:
Closes: https://github.com/h5netcdf/h5netcdf/issues/143

- [x] Tests added

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,8 +12,8 @@ Development Version:
   By `Kai Mühlbauer <https://github.com/kmuehlbauer>`_.
 - Fix regression in padding algorithm, add test.
   By `Kai Mühlbauer <https://github.com/kmuehlbauer>`_.
-- Set ``track_order=True`` by default in created files if h5py greater than
-  3.7.0 is detected to help compatibility with netCDF4-c programs.
+- Set ``track_order=True`` by default in created files if h5py 3.7.0 or 
+  greater is detected to help compatibility with netCDF4-c programs.
   By `Mark Harfouche <https://github.com/hmaarrfk>`_.
 
 Version 1.0.2 (August 2nd, 2022):
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,6 +12,9 @@ Development Version:
   By `Kai Mühlbauer <https://github.com/kmuehlbauer>`_.
 - Fix regression in padding algorithm, add test.
   By `Kai Mühlbauer <https://github.com/kmuehlbauer>`_.
+- Set ``track_order=True`` by default in created files if h5py greater than
+  3.7.0 is detected to help compatibility with netCDF4-c programs.
+  By `Mark Harfouche <https://github.com/hmaarrfk>`_.
 
 Version 1.0.2 (August 2nd, 2022):
 
--- a/h5netcdf/tests/test_h5netcdf.py
+++ b/h5netcdf/tests/test_h5netcdf.py
@@ -1176,11 +1176,7 @@ def test_reading_special_datatype_created_with_c_api(tmp_local_netcdf):
         pass
 
 
-def test_nc4_non_coord(tmp_local_netcdf):
-    # Track order True is the new default for versions after 0.12.0
-    # 0.12.0 defaults to `track_order=False`
-    # Ensure that the tests order the variables in their creation order
-    # not alphabetical order
+def test_track_order_for_variable_creation(tmp_local_netcdf):
     with h5netcdf.File(tmp_local_netcdf, "w") as f:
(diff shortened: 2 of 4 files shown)


h5netcdf/tests/test_h5netcdf.py
```diff
@@ -1191,8 +1187,21 @@ def test_nc4_non_coord(tmp_local_netcdf):
         assert f.dimensions["x"].size == 0
         assert f.dimensions["x"].isunlimited()
         assert f.dimensions["y"].size == 2
-        assert list(f.variables) == ["y", "test"]
-        assert list(f._h5group.keys()) == ["_nc4_non_coord_y", "test", "x", "y"]
```
hmaarrfk, line 1195: do you remember why we tested for `_nc4)non_coord_y`. I couldn't figure out a pattern between the two.

It seems like we shouldn't be so assertive in this case.

kmuehlbauer: `_nc4_non_coord_y` is the created variable "y", `y` is the created dimensions "y". 

We check for this, since these are two different things and we want to see if it is created correctly.

kmuehlbauer:
@hmaarrfk Thanks again for taking the lead here. I'm wondering if we should have at least one test which runs with `h5py<3.7.0`. Currently the whole matrix (beside the h5py2-tests) is running with `h5py=3.7.0`.

hmaarrfk:
I think that would be reasonable. maybe the 3.7 test? I don't really want to add complexity to the test system, but "one old dependencies" test is generally a good idea. it allows you to monitor things.

kmuehlbauer:
Yes, using the Python 3.7 test as "old" dependency test would work good enough. Would you mind adding the needed changes? 

Also the "Track Order" section in the README.rst needs update.

kmuehlbauer:
It looks like that the `provision-with-micromamba`-action does have selectors, but not the one we need. We just could add `h5py3-version` to the matrix and call for `["3.6" , 3.7"]`?

The h5netcdf builds are quite fast, maybe we can live with that for now?

hmaarrfk:
Thank you for looking for a good way to do that.

i'm somewhat doing chores now. if you have a good way to do this, feel free to push. i won't force push and we can squash merge.

kmuehlbauer:
@hmaarrfk Nice, it turns out that these h5py=3.6 tests should have been there before.

The one good thing is, we found it before release. Those errors have to be fixed before we can get along here. I hope it does not take overly long, so I assume I can fix it by tomorrow.

hmaarrfk:
how did you force push without destroying my commit history with your name?

hmaarrfk:
usually when i rebase, my little github icon appears in people's work, where im just trying to cleanup the history.

kmuehlbauer:
@hmaarrfk I've fixed the issue in #188, and dropped my earlier commit adding the h5py-matrix here. This should now work again. The last thing to do would be to add the needed changes to the README.rst "Track Order" section.

kmuehlbauer:
> how did you force push without destroying my commit history with your name?

I pulled you branch, removed my commit and force-pushed to your branch.

hmaarrfk:
i see, so you didn't rebase.

kmuehlbauer:
> i see, so you didn't rebase.

Exactly. I didn't want to break your history.

hmaarrfk:
> Exactly. I didn't want to break your history.

hmm, no issue.

What is the next version number? 1.1 or 1.0.3?

kmuehlbauer:
Depends on if we consider the recent changes more as additions than fixes. From my point of view this `track_order` would warrant a v1.1. Thoughts?

hmaarrfk:
> warrant a v1.1

Agreed.

hmaarrfk:
(fyi, i rebased and force push)

CHANGELOG.rst
```diff
@@ -12,6 +12,9 @@ Development Version:
   By `Kai Mühlbauer <https://github.com/kmuehlbauer>`_.
 - Fix regression in padding algorithm, add test.
   By `Kai Mühlbauer <https://github.com/kmuehlbauer>`_.
+- Set ``track_order=True`` by default in created files if h5py greater than
+  3.7.0 is detected to help compatibility with netCDF4-c programs.
```
kmuehlbauer, line 16:
```suggestion
- Set ``track_order=True`` by default in created files if h5py 3.7.0 or 
  greater is detected to help compatibility with netCDF4-c programs.
```

kmuehlbauer (approved):

kmuehlbauer:
@hmaarrfk Thanks you for your continuing interest and efforts in h5netcdf.

PR #192, opened 2022-11-30
Cache dtype for faster variable access

hmaarrfk:
It seems that many aspects try to access the dtype, and this can slow things down when accessing variables.

I've found that this can speed up certain access times of mine by a factor 2. I'll try to recreate it with the file I provided.

... I really wish it was easy to find small files that recreate this issue, I have to admit, the file I'm benchmarking on is about 16GB in size.  In a read only usecase.

Before

```
============= ================
--               filename 
------------- ----------------
    engine      my_file.nc
============= ================
   netcdf4      19.5±0ms      
   h5netcdf     81.2±0ms      
 h5py_direct    13.1±0ms      
============= ================
```

After
```
============= ================
--               filename 
------------- ----------------
    engine      my_file.nc
============= ================
   netcdf4      19.3±0ms      
   h5netcdf     19.9±0ms      
 h5py_direct    13.1±0ms      
============= ================
```

The benchmark looks something like:

```
import xarray as xr
dataset = xr.open_dataset(filename, engine=engine)
variable = dataset.images.variable
for i in range(30):
    variable[i].data
```

- [ ] Closes Issue #xxxx
- [ ] Tests added

### DIFF
--- a/h5netcdf/core.py
+++ b/h5netcdf/core.py
@@ -298,8 +298,9 @@ class BaseVariable(object):
                 # https://github.com/h5netcdf/h5netcdf/pull/125/
                 key = _transform_1d_boolean_indexers(key)
 
+            h5ds = self._h5ds
             if getattr(self._root, "decode_vlen_strings", False):
-                string_info = self._root._h5py.check_string_dtype(self._h5ds.dtype)
+                string_info = self._root._h5py.check_string_dtype(h5ds.dtype)
                 if string_info and string_info.length is None:
                     return self._h5ds.asstr()[key]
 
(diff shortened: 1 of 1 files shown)


h5netcdf/core.py
```diff
@@ -278,74 +279,68 @@ def __len__(self):
...
+            from .legacyapi import Dataset
 
-        if isinstance(self._parent._root, Dataset):
-            # this is only for legacyapi
-            # fix boolean indexing for affected versions
-            # https://github.com/h5py/h5py/pull/2079
-            # https://github.com/h5netcdf/h5netcdf/pull/125/
-            h5py_version = version.parse(h5py.__version__)
-            if version.parse("3.0.0") <= h5py_version < version.parse("3.7.0"):
+            if isinstance(self._parent._root, Dataset):
```
hmaarrfk, line 294: I think these accesses to the file structure are somewhat slow. I've just forked off the implementations as two separate ones.

h5netcdf/core.py
```diff
@@ -384,6 +379,40 @@ def __repr__(self):
         )
 
 
+def _get_padding(h5ds, shape, key):
```
hmaarrfk, line 382: The idea here is to really avoid creating new references to the h5ds accidentally through the `self` object. All references should be created before hand. This actually speeds things up more than I wish it would.

hmaarrfk:
Closed in favor of #197 and #198

PR #193, opened 2022-11-30
Use LRU cache to speed up access to a few read only properties

hmaarrfk:
This doesn't do anything to help the "first access" and the file load. but it does really help repeated accesses.

- [ ] Closes Issue #xxxx
- [ ] Tests added

### DIFF
--- a/h5netcdf/core.py
+++ b/h5netcdf/core.py
@@ -5,7 +5,7 @@ import warnings
 import weakref
 from collections import ChainMap, Counter, OrderedDict, defaultdict
 from collections.abc import Mapping
-from functools import lru_cache
+from functools import cached_property
 
 import h5py
 import numpy as np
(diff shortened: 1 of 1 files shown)


hmaarrfk:
TBH, I really don't want to merge this in. Doing caching at this level of granularity is quite intruisive and difficult to maintain.

I would rather understand the usecases were you think that `h5ds` can change. My understanding is that it is pretty fixed. But maybe i'm wrong.

mathause:
`lru_cache` can also make difficulties with garbage collection (search for "dont lru_cache method"). Maybe `functools.cached_property` would be better (https://docs.python.org/3.10/library/functools.html#functools.cached_property)

kmuehlbauer:
> I would rather understand the usecases were you think that h5ds can change. My understanding is that it is pretty fixed. But maybe i'm wrong.

I think there are some properties which are fixed (this compression properties seem like they won't change after creating the Dataset) and others which might change (eg shape). Maybe it would be good to make a list of properties which won't change and which might be changed?

hmaarrfk:
Closed in favor of #197 and #198

Issue #191, opened 2022-11-30
h5py minimum version update?


hmaarrfk:
Is there a minimum version for h5py that you are targeting?

It seems you are still targetting 2.10? is there a reason for this:
https://github.com/h5netcdf/h5netcdf/blob/main/.github/workflows/main.yml#L59

It seems 2.10 was released in 2019. Is that how far back you want to go?
https://pypi.org/project/h5py/2.10.0/

hmaarrfk:
FYI, according to `nep29` as far back as 3.2 should be supported:
```
$ nep29 h5py
| version |    date    |
|:-------:|:----------:|
|  3.7.0  | 2022-05-24 |
|  3.6.0  | 2021-11-16 |
|  3.5.0  | 2021-10-20 |
|  3.4.0  | 2021-08-23 |
|  3.3.0  | 2021-06-22 |
|  3.2.0  | 2021-03-03 |
```

kmuehlbauer:
@hmaarrfk I was thinking when to drop h5py2 in https://github.com/h5netcdf/h5netcdf/issues/176#issuecomment-1324646559. But maybe I'm overly concerned here and we should drop anyway? 

Maybe @shoyer has some guidance here?

shoyer:
I usually would aim to support releases from the last 1-2 years worth of releases for an upstream dependency, depending upon how stable the upstream project is and how much work it takes to support.

So I would definitely drop h5py 2.x at this point.

kmuehlbauer:
Great, thanks @shoyer! So we can get rid of the compatibility code and streamline the CI. 

That's best done in it's own PR before the performance PR's kick in. I can take care of this later today.

kmuehlbauer:
Resolved by #194, thanks @shoyer and @hmaarrfk!

PR #194, opened 2022-11-30
MNT: Remove h5py2 compatibility code

kmuehlbauer:
- [x] Closes Issue #191
- [x] Tests fixed
- [x] Changes are documented in `CHANGELOG.rst` (needs finalizing in `README.rst`)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -44,7 +44,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.8", "3.9", "3.10"]
+        python-version: ["3.8", "3.9", "3.10", "3.11"]
         h5py-version: ["3.6", "3.7"]
     steps:
     - uses: actions/checkout@v2
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -36,36 +36,6 @@ jobs:
           isort --check .
 
   build_0:
-    name: h5py2 tests
-    runs-on: ubuntu-latest
-    defaults:
-      run:
-        shell: bash -l {0}
-    strategy:
-      fail-fast: false
-      matrix:
-        python-version: ["3.7", "3.8", "3.9"]
-    steps:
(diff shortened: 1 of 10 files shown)


kmuehlbauer:
@hmaarrfk It makes much sense to use NEP29 as you suggested. I'll add a note to README.rst after sorting out the CI issues here.

hmaarrfk:
Not sure how far you want to with this. but conda-forge recently dropped python 3.7 and "nep29" would say to do the same.

kmuehlbauer:
@hmaarrfk OK, so we could even drop Python 3.7. So, let's go for it.

hmaarrfk:
I know your on it, but this also needs to get updated:
https://github.com/h5netcdf/h5netcdf/blob/main/setup.py#L36

.github/workflows/main.yml
```diff
@@ -74,7 +44,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.7", "3.8", "3.9", "3.10"]
+        python-version: ["3.8", "3.9", "3.10", "3.11"]
         h5py-version: ["3.6", "3.7"]
```
hmaarrfk, line 48:
```suggestion
        python-version: ["3.8", "3.9", "3.10"]
        h5py-version: ["3.6", "3.7"]
        include:
          - python-version: 3.11
            h5py-version: 3.7
```
https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#expanding-or-adding-matrix-configurations

kmuehlbauer:
Thanks @hmaarrfk, I've applied the suggestion!

kmuehlbauer:
@hmaarrfk I think everything is in place now, dropping h5py2 and python3.7 as well as mentioning of NEP29 and fixing CI. I'll merge by tomorrow latest, if there is nothing more to do.

kmuehlbauer:
> Sounds great! thanks for this!

Then let's get this in, so you can rebase in your performance PR's.

PR #196, opened 2022-12-01
WIP: dimension performance

kmuehlbauer:
- [ ] Closes Issue #xxxx
- [ ] Tests added


hmaarrfk:
I think this is premature optimization. We should focus on making `self._h5ds` faster.

```python
    def __len__(self):
        if self._phony:
            return self._size
        return self._h5ds.shape[0]
```

PR #197, opened 2022-12-02
Speedup accesses to h5ds by creating a hard reference to it.

hmaarrfk:
I mostly just want to run the tests.

If we can write a test that edits a dataset in a way that passes before, but fails here, I would gladly move away from this approach. But ultimately, I think it would be greatly beneficial

- [ ] Closes Issue #xxxx
- [ ] Tests added
- [ ] Changes are documented in `CHANGELOG.rst`

PR #198, opened 2022-12-03
Minor speedups to h5netcdf

hmaarrfk:
In conjunction with: 

* https://github.com/h5netcdf/h5netcdf/pull/197 <-- should be reviewed first.
* https://github.com/h5py/h5py/pull/2183

This closes the gap between h5netcdf and netcdf4 in terms of performance for some of my datasets.

Ultimately, my internal benchmarks show:

* h5netcdf suffers from increased time to access the metadata, this seems to be because accesses to `File` and dereferencing of attributes is costly. For example, if you have 6 variables, all sharing the same dimension. That dimension will be dereferenced 6 times (once for each variable) with h5netcdf. This is pretty slow.
* Once data starts getting loaded, it seems that h5netcdf is faster.
* xarray adds its own overhead, which can be prohibitive.
* 

- [ ] Closes Issue #xxxx
- [ ] Tests added

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -24,7 +24,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install  black ruff
+          pip install black ruff
       - name: Black style check
         run: |
           black --check .
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,18 +1,18 @@
 Change Log
 ----------
 
-Development Version:
+Version 1.2.0 (June 2nd, 2023):
 
 - Remove h5py2 compatibility code, remove h5py2 CI runs, mention NEP29 as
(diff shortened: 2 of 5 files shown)


h5netcdf/core.py
```diff
@@ -112,6 +113,9 @@ def __init__(self, parent, name, dimensions=None):
         self._root_ref = weakref.ref(parent._root)
         self._h5path = _join_h5paths(parent.name, name)
         self._dimensions = dimensions
+        self._h5ds = self._root._h5file[self._h5path]
+        # fix name if _nc4_non_coord_
+        self._name = self._h5path.replace("_nc4_non_coord_", "")
```
hmaarrfk, line 118: Your previous comment stated:

> dataset/name caching can be done when opened for read-only in any case, for write/append we would need to think how to apply caching/refresh cache on write

I'm not sure I know how to recreate this issue.

PR #205, opened 2022-12-13
support for HDF5 dimension scales with null dataspace

### LINKED ISSUE #204: support for HDF5 dimension scales with null dataspace
#### Is your feature request related to a problem? Please describe.

I cannot use `h5netcdf` as the `XArray.open_dataset` engine to read HDF5 files that contain a dimension scale with null dataset.

#### Describe the solution you'd like

A dimension scale with null dataset to be handled as a netCDF dimension but not a variable.

#### Describe alternatives you've considered

The `netcdf4-python` package also fails in this case, more spectacularly than `h5netcdf`, by crashing Python. I've raised Unidata/netcdf4-python#1226, but am unable to offer a PR because the problem is in the netcdf4-c library.

#### Additional context

I'm am submitting a PR that closes this issue.

itcarroll:
- [x] Closes #204
- [x] Tests added

This PR adds support for HDF5 dimension scales that have a null dataset, which are handled as netCDF dimensions but not netCDF variables. I'll push an update to the CHANGELOG if you support this enhancement and I've responded to any changes you would like.

Currently h5netcdf throws this error when encountering a dimension scale with null dataset:

```
  File "h5netcdf/h5netcdf/dimensions.py", line 232, in __len__
    return self._h5ds.shape[0]
TypeError: 'NoneType' object is not subscriptable
```
The `NoneType` is expected on a dimension scale with null dataset. The core of my PR is a change to the `Dimension.size` property:

```
@@ -108,7 +110,24 @@ class Dimension(object):
     @property
     def size(self):
         """Return dimension size."""
-        size = len(self)
+        vars = self._parent._h5group["/"]
+        if not self._phony and self._h5ds.shape is None:
+            # the dimension is a null dataset, essentially just a shared name,
+            # and is only compatible with netCDF4 if all axes that reference
+            # the dimension have the same length
+            try:
+                reflist = self._h5ds.attrs["REFERENCE_LIST"]
+            except KeyError:
+                msg = f"Dimension {self.name!r} has no discernable length."
+                raise CompatibilityError(msg)
+            ref_size = {vars[ref].shape[axis] for ref, axis in reflist}
+            try:
+                (size, ) = ref_size
+            except ValueError:
+                msg = f"Dimension {self.name!r} references axes of unequal length."
+                raise CompatibilityError(msg)
+        else:
+            size = len(self)
```

However, changes to the following were required to ensure all tests passed.
- `_netcdf_dimension_but_not_variable`
- `CompatibilityError` moved to h5netcdf.utils
- `Group` initialization tests for adding variables

### DIFF
--- a/h5netcdf/core.py
+++ b/h5netcdf/core.py
@@ -13,7 +13,7 @@ from packaging import version
 from . import __version__
 from .attrs import Attributes
 from .dimensions import Dimension, Dimensions
-from .utils import Frozen
+from .utils import Frozen, CompatibilityError
 
 try:
     import h5pyd
@@ -34,10 +34,6 @@ def _name_from_dimension(dim):
     # First value in a dimension is the actual dimension scale
     # which we'll use to extract the name.
     return dim[0].name.split("/")[-1]
-
-
-class CompatibilityError(Exception):
-    """Raised when using features that are not part of the NetCDF4 API."""
 
 
 def _invalid_netcdf_feature(feature, allow):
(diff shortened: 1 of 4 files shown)


kmuehlbauer:
@itcarroll I'll close this PR for the time being, until there is an actual plan of action. If netcdf-c has finally decided how to approach that issue one way or the other I'm happy to reiterate. We'll keep #204 open. Thanks for the patience so far!

PR #220, opened 2023-10-23
Allow to use ros3 driver for remote url

zequihg50:
Currently, for remote URLs `hpy5d` is used. However, it is desirable to allow the usage of the `ros3` HDF5 driver to read from remote URLs.

The following code does not work at the moment.

```python
import h5netcdf

f = h5netcdf.File("https://api.cloud.ifca.es:8080/swift/v1/IPCC-IA-Monthly/test2.nc", driver="ros3")
print(list(f))
```

This simple pull request allows to use the `ros3` driver with `h5py` ignoring `h5pyd`.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,11 @@
 Change Log
 ----------
 
+Development Version:
+
+- Add ros3 support by checking `driver`-kwarg.
+  By `Ezequiel Cimadevilla Alvarez <https://github.com/zequihg50>`_
+
 Version 1.2.0 (June 2nd, 2023):
 
 - Remove h5py2 compatibility code, remove h5py2 CI runs, mention NEP29 as
--- a/h5netcdf/core.py
+++ b/h5netcdf/core.py
@@ -1029,7 +1029,10 @@ class File(Group):
         self.decode_vlen_strings = kwargs.pop("decode_vlen_strings", None)
         try:
             if isinstance(path, str):
-                if path.startswith(("http://", "https://", "hdf5://")) and "driver" not in kwargs:
+                if (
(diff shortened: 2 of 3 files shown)


kmuehlbauer:
Sory for the delay here.

@zequihg50 Is there an endpoint to test this? 

I'm hesitant to merge without at least a local non-CI test.

zequihg50:
Maybe [this one](https://www.unidata.ucar.edu/software/netcdf/examples/OMI-Aura_L2-example.nc) from the netCDF examples?

kmuehlbauer:
Thanks @zequihg50 for coming back. To test the ros3 access the file would have to be in an aws bucket, or am I mistaken? 

OTOH, we could just assume that ros3 access was tested properly over at h5py and get away with it here.

zequihg50:
It works by issuing http range requests so any server supporting them it is valid. I have tested the netCDF URL with the ros3 driver and it works.

kmuehlbauer:
@zequihg50 Do you feel like adding a test? If not I can take it from here. Just let me know.

zequihg50:
I have added a very basic test just to check that the file opens correctly but feel free to edit it in anyway you consider appropriate. Just mention that this requires h5py built with ros3 support (I think installation from pip does not include ros3 support).

kmuehlbauer:
I'll check and add necessary code and information and merge afterwards. Thanks @zequihg50!

kmuehlbauer (approved):

kmuehlbauer:
Thanks again @zequihg50! I'll check and can get a release out early next week latest.

zequihg50:
Awesome, thanks!

Issue #195, opened 2022-11-30
Improving performance for h5netcdf


hmaarrfk:
#### What happened:

I've been trying to use h5netcdf since it gives access to the underlying `h5py` dataset which makes it easier to prototype with different HDF5 backends. netcdf4-c is much harder to compile and thus prototype with since many of the internals are not accessible.

My original concern arose from observations that, for some of our datasets, they open with

  * xarray + netcdf4-python in about 80 ms
  * xarray + h5netcdf: 600 ms.

Ultimately, I believe that I think things boil down to the fact an access to the `_h5ds` property is quite costly. This makes "every" operation "slow".

the code below:
```python
import h5netcdf
from pathlib import Path

h5nc_file = h5netcdf.File(
    Path.home() / 'sample_file_zebrafish_0.18.55_4uCams_small.nc',
)

images = h5nc_file.variables['images']
for i in range(1000):
    images.shape
```

Using the spyder profile, one finds that 1000 calls to size, result in about 8000 calls to `_h5ds`
![image](https://user-images.githubusercontent.com/90008/204885209-45f6749b-faf9-432d-8ce0-522585dbe686.png)

```python
import h5netcdf
from pathlib import Path

h5nc_file = h5netcdf.File(
    Path.home() / 'sample_file_zebrafish_0.18.55_4uCams_small.nc',
)

images = h5nc_file.variables['images']

for i in range(8000):
    images._h5ds
```

This is because this property is created on demand.

I think this has cascading effects because everything really relies on the access to the underlying `h5py` structure.

A previous suggestion tried to remove the property, but it seems like there was some concern about "resetting" the underlying `_h5py` pointer. I am unclear on where this kind of usecase would come up. My understanding is that this pointer would be something unchanging.

#### Expected Output
As a comparison, we can compare how long it takes to access the shape property from netcdf4:
```
import h5netcdf
from pathlib import Path
import netCDF4

h5nc_file = netCDF4.Dataset(
    str(Path.home() / 'sample_file_zebrafish_0.18.55_4uCams_small.nc'),
)

images = h5nc_file.variables['images']
for i in range(1000):
    images.shape
```

About 400 us instead of 350 ms.
![image](https://user-images.githubusercontent.com/90008/204886355-5dc90407-1152-4408-8714-c74c95993ec9.png)

#### Anything else we need to know?:
The sample image has 4 dimensions

[...]

PR #221, opened 2023-11-03
Use cached properties to speed up dataset access

pnuu:
My use case involves using `h5netcdf` to read geostationary satellite data in the Satpy library. The data, new geostationary Meteosat FCI instrument by EUMETSAT, has a very complicated NetCDF4 structure split over 80 separate files. The complicated structure prevents us from using `xr.open_dataset()` as we'd end up having tens of thousands of open file handles.

For local reading we are using `netCDF4-python`, but to get remote access (e.g. from a local CEPH/S3) via `fsspec` we are using `h5netcdf`. Just collecting the metadata from local files (creating a `Scene` object in Satpy) using `h5netcdf` takes 36 seconds. As a reference, `netCDF4-python` also uses 10 seconds for this, which already tells how complex the files are.

The discussion in #195 and the work on speeding up in #197 has been of great value when trying to understand the causes. 

Inspired by these I had a look, and added some caching for some properties in `h5netcdf.core` and 10 seconds from the execution time with the tradeoff of using 1100 MB of RAM instead of 800 MB using the current `main`.

The biggest gain came from caching the `BaseVariable._h5ds` property.

I could get still few seconds of caching more, but then I encounter test failure (in `h5netcdf/tests/test_h5netcdf.py::test_no_circular_references` having 2 references instead of the 1 asserted in the tests) that I didn't want to tweak.

- [ ] Closes Issue #xxxx
- [ ] Tests added

### DIFF
--- a/h5netcdf/core.py
+++ b/h5netcdf/core.py
@@ -548,7 +548,7 @@ class Group(Mapping):
     def _parent(self):
         return self._parent_ref()
 
-    @property
+    @cached_property
     def _h5group(self):
         # Always refer to the root file and store not h5py object
         # subclasses:
--- a/h5netcdf/core.py
+++ b/h5netcdf/core.py
@@ -109,18 +109,19 @@ def _expanded_indexer(key, ndim):
 
 class BaseVariable:
     def __init__(self, parent, name, dimensions=None):
-        self._parent = parent
+        self._parent_ref = weakref.ref(parent)
+        self._root_ref = weakref.ref(parent._root)
         self._h5path = _join_h5paths(parent.name, name)
         self._dimensions = dimensions
         self._initialized = True
 
     @cached_property
     def _parent(self):
-        return self._parent
+        return self._parent_ref()
 
     @cached_property
     def _root(self):
-        return self._parent._root
+        return self._root_ref()
 
     @cached_property
     def _h5ds(self):
@@ -411,7 +412,7 @@ class _LazyObjectLookup(Mapping):
         self._object_cls = object_cls
         self._objects = OrderedDict()
 
-    @property
+    @cached_property
     def _parent(self):
         return self._parent_ref()
 
--- a/h5netcdf/core.py
+++ b/h5netcdf/core.py
@@ -109,19 +109,18 @@ def _expanded_indexer(key, ndim):
 
 class BaseVariable:
     def __init__(self, parent, name, dimensions=None):
-        self._parent_ref = weakref.ref(parent)
-        self._root_ref = weakref.ref(parent._root)
+        self._parent = parent
         self._h5path = _join_h5paths(parent.name, name)
         self._dimensions = dimensions
         self._initialized = True
 
-    @property
+    @cached_property
     def _parent(self):
-        return self._parent_ref()
+        return self._parent
 
-    @property
+    @cached_property
     def _root(self):
-        return self._root_ref()
+        return self._parent._root
 
     @cached_property
     def _h5ds(self):
--- a/h5netcdf/core.py
+++ b/h5netcdf/core.py
@@ -1068,7 +1068,6 @@ class File(Group):
 
         self._mode = mode
         self._writable = mode != "r"
-        self._root_ref = weakref.ref(self)
         self._h5path = "/"
         self.invalid_netcdf = invalid_netcdf
 
--- a/h5netcdf/core.py
+++ b/h5netcdf/core.py
@@ -128,7 +128,7 @@ class BaseVariable:
         # subclasses:
         return self._root._h5file[self._h5path]
 
-    @property
+    @cached_property
     def name(self):
         """Return variable name."""
         # fix name if _nc4_non_coord_
@@ -253,7 +253,7 @@ class BaseVariable:
         if self._h5ds.shape != new_shape:
             self._h5ds.resize(new_shape)
 
-    @property
+    @cached_property
     def dimensions(self):
         """Return variable dimension names."""
         if self._dimensions is None:
@@ -266,7 +266,7 @@ class BaseVariable:
         # return actual dimensions sizes, this is in line with netcdf4-python
         return tuple([self._parent._all_dimensions[d].size for d in self.dimensions])
 
-    @property
+    @cached_property
     def ndim(self):
         """Return number variable dimensions"""
         return len(self.shape)
@@ -274,7 +274,7 @@ class BaseVariable:
     def __len__(self):
         return self.shape[0]
 
-    @property
+    @cached_property
     def dtype(self):
         """Return NumPy dtype object giving the variable’s type."""
         return self._h5ds.dtype
@@ -357,7 +357,7 @@ class BaseVariable:
             self._maybe_resize_dimensions(key, value)
         self._h5ds[key] = value
 
-    @property
+    @cached_property
     def attrs(self):
         """Return variable attributes."""
         return Attributes(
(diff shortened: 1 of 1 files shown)


kmuehlbauer:
@pnuu Thanks for sharing your use case and this PR. As you already pointed out the whole issue is quite intricate when it comes to circular references. Another tricky issue is keeping references to underlying hdf5 dataset's which might break downstream (eg. xarray) code. 

@hmaarrfk has a much better understanding and I hope he can share thoughts on that.

pnuu:
This is all the properties that can be cached without breaking the existing tests, and also giving me so far the best performance. Some of the remaining properties cause failure in circular dependency test, and others `HDF Error` in the legacy API.

I'm trying to figure out why the legacy tests fail if I cache `Dimension._h5ds` (which would give me another few second speedup), but first of all can't figure out where the `read_module` fixture is coming from.

kmuehlbauer:
`read_module` is actually no real fixture, it's _defined_ in this helper function, which will be called with the real test-functions:

https://github.com/h5netcdf/h5netcdf/blob/2771793b27588026bf11b81b9cd77aabec59e187/h5netcdf/tests/test_h5netcdf.py#L446-L449

pnuu:
> `read_module` is actually no real fixture, it's _defined_ in this helper function, which will be called with the real test-functions:
> 
> https://github.com/h5netcdf/h5netcdf/blob/2771793b27588026bf11b81b9cd77aabec59e187/h5netcdf/tests/test_h5netcdf.py#L446-L449

Oooh, indeed, thanks!

kmuehlbauer:
@pnuu I've relaxed the restriction on running actions for new contributors. Now you should be able to push and the CI will directly kick in.

pnuu:
Thank you @kmuehlbauer !

I can get the legacy API tests to pass while caching `Dimension._hd5s` and `Dimension._Dimension._group` for `mode="r"` (and again gain few seconds in processing time, now down to 21.6 s), but then get failures in `test_h5netcdf.py::test_no_circular_references`. I'll have a look at that before pushing the latest cache additions.

pnuu:
I might be misunderstanding a lot, but I'm not sure the `test_no_circular_references()` tests for _circular_ references, just that there should only be one reference to `ds` while the file is open. The referrer count goes up even with `print(ds)` which isn't very circular. The [documentation for `.get_referrers()`](https://docs.python.org/3/library/gc.html#gc.get_referrers) talks only about direct references.

Another thing I was thinking is that it might be safer to do caching only for `mode="r"`, even if the tests are currently passing. Thougths? This will need quite a bit refactoring.

hmaarrfk:
i think this is also the source of test failures for xarray.

We need to make it a weakref somehow. 

One reason I didn't use `cache_property` decorator in the past is that it is hard to "invalidate" and reset. I typically roll my own with a dictionary.

hmaarrfk:
Personally, the performance drawbacks of xarray and netcdf have made me just move away from coordinates all together.

I've decided to place most things in "data" variables. quality of life goes way up....

pnuu:
> i think this is also the source of test failures for xarray.
> 
> We need to make it a weakref somehow.

What are you referring to in particular in these?

> One reason I didn't use `cache_property` decorator in the past is that it is hard to "invalidate" and reset. I typically roll my own with a dictionary.

I already have one branch on my work laptop that has a helper method for cache invalidation for `@cached_property` decorated properties. I'll see how ugly the "cache only if `mode="r"`" variant becomes and add the cache invalidation after that.

hmaarrfk:
See what I tried in: https://github.com/h5netcdf/h5netcdf/pull/197/files#diff-8ccad2bb9409c656f6ea28ec8e387bd01f301ef429355d60a837a90d9aaa316dR117

lahtinep:
Caching now only happens for `mode="r"` and only for cases where the access is more than a simple attribute access. I also added some tests for the cache handling and clearing.

The next I think I'll try to find some reasoning what `test_no_circular_references` is really about. I don't see why there shouldn't be more than one _direct_ reference to `ds` while the file is still open.

lahtinep:
Damn, I try to clean `File._groups` which doesn't exist (`File` inherits from `Group`, but doesn't call its `__init__()`). I need to check tomorrow how the groups are tied to the `File` class.

kmuehlbauer:
The call to super is after the complete initialization step.

    https://github.com/h5netcdf/h5netcdf/blob/2771793b27588026bf11b81b9cd77aabec59e187/h5netcdf/core.py#L1102

pnuu:
I think I'm ready with this stage of optimization.

I've adapted the test script from https://github.com/h5netcdf/h5netcdf/issues/195#issuecomment-1336182157 to use caching benefits of `mode="r"`. The file is also closed between the tests so that the caching from the previous steps don't affect the test.

```python

import h5netcdf
from pathlib import Path
from time import perf_counter, sleep

import tempfile

NUM = 10000

with tempfile.TemporaryDirectory() as tmpdir:
    filename = Path(tmpdir) / 'h5netcdf_test.nc'

    with h5netcdf.File(filename, 'w') as h5nc_file:
        h5nc_file.dimensions = {'x': 1024}

    with h5netcdf.File(filename, 'r') as h5nc_file:
        x = h5nc_file.dimensions['x']
        time_start = perf_counter()
        for i in range(NUM):
            len(x)
        time_end = perf_counter()
        time_elapsed = time_end - time_start
        print(f"{time_elapsed:03.3f} --- Time to compute length {NUM} times: ")
        
    with h5netcdf.File(filename, 'r') as h5nc_file:
        x = h5nc_file.dimensions['x']
        time_start = perf_counter()
        for i in range(NUM):
            x._h5ds
        time_end = perf_counter()
        time_elapsed = time_end - time_start
        print(f"{time_elapsed:03.3f} --- Time to access h5ds reference {NUM} times")
    
    with h5netcdf.File(filename, 'r') as h5nc_file:
        x = h5nc_file.dimensions['x']
        time_start = perf_counter()
        for i in range(NUM):
            x._root._h5file[x._h5path]
        time_end = perf_counter()
        time_elapsed = time_end - time_start
        print(f"{time_elapsed:03.3f} --- Time to create h5ds reference {NUM} times")
```

Running this with `main` gives me

```
$ python test_h5netcdf.py 
0.711 --- Time to compute length 10000 times: 
0.602 --- Time to access h5ds reference 10000 times
0.606 --- Time to create h5ds reference 10000 times
```

and with this PR

```$ python test_h5netcdf.py 
0.007 --- Time to compute length 10000 times: 
0.003 --- Time to access h5ds reference 10000 times
0.612 --- Time to create h5ds reference 10000 times
```

Note that the last of these tests is the same as the second one without using the readily available properties so the caching doesn't help. I've also done 10x the accesses compared to #195 to have something non-zero as a result.

My own use case went from 34 s to 23 s with these changes.

veenstrajelmer:
I have installed this branch to see if it speeds up opening my datasets. My issue is documented in https://github.com/Deltares/dfm_tools/issues/968, which shows that `xr.open_mfdataset()` is way slower with `engine="h5netcdf"` compared to `engine="netcdf4"` (but consumes much less memory). I noticed this is caused by calls to `name()` as documented in https://github.com/h5netcdf/h5netcdf/issues/195#issuecomment-2302019051. I see this branch reduces the calls from 7303 to 4873, but it increases the time per call from 0.003/0.004 sec to 0.0053 sec, which ends up consuming as much or more time. Not sure if my issue is relevant for this fix, but I thought it would be worth reporting it either way. Please note my dataset has 2410 variables, which seems quite sub-optimal. I assume `name()` is called for all these variables and maybe also their dimensions.

PR #223, opened 2023-11-10
Prepare for inclusion of driver_kwds in xarray

zequihg50:
Prepare for the inclusion of `driver_kwds` in [xarray](https://github.com/pydata/xarray/pull/8360#discussion_r1387568762).

- [X] Tests added
- [] Changes are documented in `CHANGELOG.rst`

### DIFF
--- a/h5netcdf/core.py
+++ b/h5netcdf/core.py
@@ -1049,6 +1049,11 @@ class File(Group):
                         path, mode, track_order=track_order, **kwargs
                     )
                 else:
+                    driver_kws = kwargs.pop("driver_kwds", dict())
+                    if driver_kws is None:
+                        driver_kws = dict()
+                    kwargs.update(driver_kws)
+
                     self._preexisting_file = os.path.exists(path) and mode != "w"
                     self._h5py = h5py
                     self._h5file = self._h5py.File(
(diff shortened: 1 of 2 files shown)


zequihg50:
I'm not sure if `driver_kwds` is going to work in xarray due to [CachingFileManager](https://github.com/pydata/xarray/blob/e5d163a8ddef4e8aa95a2841505a642f8ffbcea6/xarray/backends/h5netcdf_.py#L174) and `driver_kwds` not being hashable, maybe it is better to unpack the `driver_kwds` in multiple parameters.

kmuehlbauer:
> I'm not sure if `driver_kwds` is going to work in xarray due to [CachingFileManager](https://github.com/pydata/xarray/blob/e5d163a8ddef4e8aa95a2841505a642f8ffbcea6/xarray/backends/h5netcdf_.py#L174) and `driver_kwds` not being hashable, maybe it is better to unpack the `driver_kwds` in multiple parameters.

@zequihg50 I completely agree, we should just use the keywords which h5py is expecting. Then we can do unpacking `driver_kwds` over at xarray and just feed this here via `kwargs` without having to handle them here.

Then we do not have to handle prepare anything here, right?

zequihg50:
Ok, I see. Then there is nothing to add here. I will keep this open just in case but I will close after all is merged in xarray.

zequihg50:
Closing, no need for this, see https://github.com/pydata/xarray/pull/8360.

PR #227, opened 2024-01-18
add VLType and CompoundType, commit complex compound type to file

kmuehlbauer:
- [ ] Closes Issue #xxxx
- [x] Tests added
- [x] fix char array issue for compound type

see #228 for transient vs committed types

### DIFF
--- a/h5netcdf/core.py
+++ b/h5netcdf/core.py
@@ -173,7 +173,7 @@ class UserType(BaseObject):
 
     @property
     def name(self):
-        """Return enum name."""
+        """Return user type name."""
         # strip hdf5 path
         return super().name.split("/")[-1]
 
@@ -211,6 +211,7 @@ class EnumType(UserType):
 
     @property
     def enum_dict(self):
+        """Dictionary containing the Enum field/value pairs."""
         return self.dtype.metadata["enum"]
 
     def __repr__(self):
--- a/h5netcdf/core.py
+++ b/h5netcdf/core.py
@@ -354,14 +354,23 @@ class BaseVariable(BaseObject):
     @property
     def datatype(self):
         """Return numpy dtype or user defined type."""
-        if (enum_dict := self._root._h5py.check_enum_dtype(self.dtype)) is not None:
-            for tid in self._parent._all_enumtypes.values():
-                if self._root._h5py == h5py:
-                    found = self._h5ds.id.get_type().equal(tid._h5ds.id)
-                else:
-                    found = tid.enum_dict == enum_dict
-                if found:
+        # this is really painful as we have to iterate over all types
+        # and check equality
+        usertype = None
+        metadata = self.dtype.metadata if self.dtype.metadata else {}
+        if "enum" in metadata:
+            usertype = self._parent._all_enumtypes
+        elif "vlen" in metadata:
+            usertype = self._parent._all_vltypes
+        elif self.dtype.names is not None or "complex" in self.dtype.name:
+            usertype = self._parent._all_cmptypes
+
+        if usertype is not None:
+            for tid in usertype.values():
+                if self.dtype == tid.dtype and metadata == tid.dtype.metadata:
                     return tid
+
+        # fallback to just dtype
         return self.dtype
 
     def _get_padding(self, key):
--- a/h5netcdf/core.py
+++ b/h5netcdf/core.py
@@ -223,7 +223,7 @@ class VLType(UserType):
 
 def _string_to_char_array_dtype(dtype):
     """Converts fixed string to char array dtype."""
-    if np.isdtype(dtype, "complex floating"):
+    if dtype.kind == "c":
         return None
     return np.dtype(
         {
@@ -238,7 +238,7 @@ def _string_to_char_array_dtype(dtype):
 
 def _char_array_to_string_dtype(dtype):
     """Converts char array to fixed string dtype."""
-    if np.isdtype(dtype, "complex floating"):
+    if dtype.kind == "c":
         return None
     return np.dtype(
         {
--- a/h5netcdf/core.py
+++ b/h5netcdf/core.py
@@ -139,7 +139,7 @@ class BaseObject:
         return self._h5ds.dtype
 
 
-h5type_mapping = {
+_h5type_mapping = {
     "H5T_COMPOUND": 6,
     "H5T_ENUM": 8,
     "H5T_VLEN": 9,
@@ -147,19 +147,25 @@ h5type_mapping = {
 
 
 def _get_h5usertype_identifier(h5type):
+    """Return H5 Type Identifier from given H5 Datatype."""
     try:
-        tkey = h5type.id.get_class()
+        # h5py first
+        h5typeid = h5type.id.get_class()
     except AttributeError:
-        tkey = h5type_mapping[h5type.id.type_json["class"]]
-    return tkey
+        # h5pyd second
+        h5typeid = _h5type_mapping[h5type.id.type_json["class"]]
+    return h5typeid
 
 
 def _get_h5dstype_identifier(h5type):
+    """Return H5 Type Identifier from given H5 Dataset."""
     try:
-        tkey = h5type.id.get_type().get_class()
+        # h5py first
+        h5typeid = h5type.id.get_type().get_class()
     except AttributeError:
-        tkey = h5type_mapping[h5type.id.type_json["class"]]
-    return tkey
+        # h5pyd second
+        h5typeid = _h5type_mapping[h5type.id.type_json["class"]]
+    return h5typeid
 
 
 class UserType(BaseObject):
@@ -355,7 +361,11 @@ class BaseVariable(BaseObject):
             self._h5ds.resize(new_shape)
 
     def _add_fillvalue(self, fillvalue):
-        """Add _FillValue attribute"""
+        """Add _FillValue attribute.
+
+        This method takes care of adding fillvalue with the wanted
+        variable dtype.
+        """
 
         # trying to create correct type of fillvalue
         if self.dtype is str:
@@ -409,6 +419,9 @@ class BaseVariable(BaseObject):
     @property
     def _h5datatype(self):
         """Returns comparable h5type.
+
+        This property can be used to compare two variables/datatypes or
+        a variable and a datatype for equality of the underlying datatype.
 
         - DatatypeID for h5py
         - (dtype, dtype.metadata) for h5pyd
@@ -626,20 +639,23 @@ def _unlabeled_dimension_mix(h5py_dataset):
     return status
 
 
-def _check_dtype(self, dtype):
-    """Check and handle dtypes"""
+def _check_dtype(group, dtype):
+    """Check and handle dtypes when adding variable to given group.
+
+    Raises errors and issues warnings according to given dtype.
+    """
 
     if dtype == np.bool_:
         # never warn since h5netcdf has always errored here
         _invalid_netcdf_feature(
             "boolean dtypes",
-            self._root.invalid_netcdf,
+            group._root.invalid_netcdf,
         )
     else:
-        self._root._check_valid_netcdf_dtype(dtype)
+        group._root._check_valid_netcdf_dtype(dtype)
 
     # we only allow h5netcdf user types, not named h5py.Datatype
-    if isinstance(dtype, self._root._h5py.Datatype):
+    if isinstance(dtype, group._root._h5py.Datatype):
         raise TypeError(
             f"Argument dtype {dtype!r} is not allowed. "
             f"Please provide h5netcdf user type or numpy compatible type."
@@ -649,18 +665,18 @@ def _check_dtype(self, dtype):
     # we just use the h5py user type here
     if isinstance(dtype, (EnumType, VLType, CompoundType)):
         h5type = dtype._h5ds
-        if dtype._root._h5file.filename != self._root._h5file.filename:
+        if dtype._root._h5file.filename != group._root._h5file.filename:
             raise TypeError(
                 f"Given dtype {dtype} is not committed into current file"
-                f" {self._root._h5file.filename}. Instead it's committed into"
+                f" {group._root._h5file.filename}. Instead it's committed into"
                 f" file {dtype._root._h5file.filename}"
             )
         # check if committed type can be accessed in current group hierarchy
-        user_type = self._get_usertype(h5type)
+        user_type = group._get_usertype(h5type)
         if user_type is None:
             msg = (
                 f"Given dtype {dtype.name!r} is not accessible in current group"
-                f" {self._h5group.name!r} or any parent group. Instead it's defined at"
+                f" {group._h5group.name!r} or any parent group. Instead it's defined at"
                 f" {h5type.name!r}. Please create it in the current or any parent group."
             )
             raise TypeError(msg)
@@ -685,26 +701,30 @@ def _check_dtype(self, dtype):
         #  which may be used here
         # if dname is not available in current group-path
         # create and commit type in current group
-        if dname not in self._all_cmptypes:
-            dtype = self.create_cmptype(dtype, dname).dtype
+        if dname not in group._all_cmptypes:
+            dtype = group.create_cmptype(dtype, dname).dtype
 
     return dtype
 
 
-def _check_fillvalue(self, fillvalue, dtype):
-    """Handles fillvalues before dataset creation"""
+def _check_fillvalue(group, fillvalue, dtype):
+    """Handles fillvalues when adding variable to given group.
+
+    Raises errors and issues warnings according to
+    given fillvalue and dtype.
+    """
 
     # handling default fillvalues for legacyapi
     # see https://github.com/h5netcdf/h5netcdf/issues/182
     from .legacyapi import Dataset, _get_default_fillvalue
 
-    stacklevel = 5 if isinstance(self._parent, Dataset) else 4
+    stacklevel = 5 if isinstance(group._root, Dataset) else 4
 
     h5fillvalue = fillvalue
 
     # if no fillvalue is provided take netcdf4 default values for legacyapi
     if fillvalue is None:
-        if isinstance(self._parent, Dataset):
+        if isinstance(group._root, Dataset):
             h5fillvalue = _get_default_fillvalue(dtype)
 
     # handling for EnumType
@@ -1104,7 +1124,7 @@ class Group(Mapping):
         dimensions : tuple
             Tuple containing dimension name strings. Defaults to empty tuple, effectively
             creating a scalar variable.
-        dtype : numpy.dtype, str, optional
+        dtype : numpy.dtype, str, UserType (Enum, VL, Compound), optional
             Datatype of the new variable. Defaults to None.
         fillvalue : scalar, optional
             Specify fillvalue for uninitialized parts of the variable. Defaults to ``None``.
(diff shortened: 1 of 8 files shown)


kmuehlbauer:
@ZedThree FYI, this is the first serve of adding the missing user types (committed types) to h5netcdf as well as making it compatible with nc-complex.

h5netcdf/core.py
```diff
@@ -724,6 +748,21 @@ def _create_child_variable(
...
+        # compound type handling
+        if np.dtype(dtype).kind == "c":
+            if not isinstance(dtype, CompoundType) and self._root._auto_complex:
+                # todo check compound types for
+                #  existing complex types which may be used
+                itemsize = np.dtype(dtype).itemsize
+                if itemsize == 8:
+                    width = "FLOAT"
+                elif itemsize == 16:
+                    width = "DOUBLE"
```
ZedThree, line 760: I don't know how common other type sizes are, but it's probably wise to explicitly error here

ZedThree:
LGTM! I think your xarray tests are failing because, somewhat ironically, they actually work and you expect them to fail! :smile: 

Just to note that nc-complex supports a wider variety of conventions for complex numbers, but I think it's absolutely fine for h5netcdf to only support the h5py/transient compound type and the explicit `_PFNC_*_COMPLEX` named type. h5netcdf will be able to read them all perfectly ok, just not auto-detect them as being complex numbers.

kmuehlbauer:
Thanks for taking a look here, much appreciated.

> LGTM! I think your xarray tests are failing because, somewhat ironically, they actually work and you expect them to fail! 😄

Yes, they are actually from xarray's own test-suite, which expect h5netcdf to error here :grinning: 

> Just to note that nc-complex supports a wider variety of conventions for complex numbers, but I think it's absolutely fine for h5netcdf to only support the h5py/transient compound type and the explicit `_PFNC_*_COMPLEX` named type. h5netcdf will be able to read them all perfectly ok, just not auto-detect them as being complex numbers.

Sure, we can think about aligning h5netcdf's legacyapi to what netcdf4-python supports.

ZedThree:
> Sure, we can think about aligning h5netcdf's legacyapi to what netcdf4-python supports.

This would be neat, but much less important than your work here to ensure that currently released (and older) versions of netcdf will be able to read files created by h5netcdf. This is a great improvement for compatibility, so thank you!

kmuehlbauer:
@shoyer, I'd like to merge this in the coming days. If you have time it would be great if you could take a look here. Thanks!

With the work of @ZedThree over in https://github.com/PlasmaFAIR/nc-complex and this PR here, we can enable complex numbers for xarray netcdf4/h5netcdf backends.

h5netcdf/core.py
shoyer: `width = {8: "FLOAT", 16: "DOUBLE"}[itemsize]` still feels like a safer way to write this tp me :)

h5netcdf/core.py
```diff
@@ -1127,6 +1211,7 @@ def __init__(self, path, mode="r", invalid_netcdf=False, phony_dims=None, **kwar
         track_order = kwargs.pop("track_order", track_order_default)
 
         self.decode_vlen_strings = kwargs.pop("decode_vlen_strings", None)
+        self._auto_complex = kwargs.pop("auto_complex", None)
```
shoyer, line 1214: Why not use real keyword arguments here (which could be documented in the docstring) instead of putting them in `**kwargs`?

h5netcdf/core.py
```diff
@@ -1192,9 +1277,17 @@ def __init__(self, path, mode="r", invalid_netcdf=False, phony_dims=None, **kwar
...
+            # this creates files which are readable by netCDF4 with nc-complex
+            # as well as h5netcdf
+            if self._auto_complex is None:
+                self._auto_complex = True
         else:
             if self.decode_vlen_strings is None:
                 self.decode_vlen_strings = False
+            # default _auto_complex to False for new API
+            if self._auto_complex is None:
+                self._auto_complex = False
```
shoyer, line 1290: This mismatch between the legacy and new APIs is a little strange and could surprise users. A more conservative choice (which could be changed later) is to defautl to `auto_complex=False` and require users to explicitly opt-in to the new behavior.

Are arrays written with `auto_complex=True` automatically read as complex valued with h5py or old versions of h5netcdf? If not, I would stick with `auto_complex=False` as a default in both cases until a concensus emerges that this is the right way to handle complex values in netCDF files.

h5netcdf/tests/test_h5netcdf.py
```diff
@@ -2404,6 +2404,181 @@ def test_enum_type_creation(tmp_local_or_remote_netcdf, netcdf_write_module):
...
+    complex_array = np.array([0 + 0j, 1 + 0j, 0 + 1j, 1 + 1j, 0.25 + 0.75j])
+    kwargs = dict(auto_complex=True)
+
+    with netcdf_write_module.Dataset(tmp_local_netcdf, "w", **kwargs) as ds:
+        ds.createDimension("x", size=len(complex_array))
+        if dtype == "c4":
+            try:
+                var = ds.createVariable("data", dtype, ("x",))
+            except TypeError as e:
+                assert e.args[0] == "data type 'c4' not understood"
```
shoyer, line 2564: pytest.raises is the standard way to write this: https://seaborn.pydata.org/tutorial/color_palettes.html#perceptually-uniform-palettes

h5netcdf/tests/test_h5netcdf.py
```diff
@@ -2404,6 +2404,181 @@ def test_enum_type_creation(tmp_local_or_remote_netcdf, netcdf_write_module):
...
+                if netcdf_write_module.__name__ == "netCDF4":
+                    assert "Illegal primitive data type" in e.args[0]
+                else:
+                    assert (
+                        e.args[0]
+                        == "Currently only 'complex64' and 'complex128' dtypes are allowed."
+                    )
+        else:
+            var = ds.createVariable("data", dtype, ("x",))
+            var[:] = complex_array
```
shoyer, line 2579: Generally it's cleaner to write separate unit tests rather than a parameterized test with branched logic.

h5netcdf/core.py
kmuehlbauer: I was hesitant to do this. But it will surely be more visible, thanks!

h5netcdf/core.py
kmuehlbauer: The only difference for h5py/h5netcdf is the additional committed (named) type within the file on write. Old versions can read such files, relying on the type retrieved from the dataset. New versions of h5netcdf also gather the type in `.cmptypes`.

But going conservative first is a good idea anyway. netcdf4-python does the same, @ZedThree, right?

Thanks Stefan for the review, much appreciated.  😊

kmuehlbauer:
I could not get around to check this in more detail in this gist: https://gist.github.com/kmuehlbauer/1f6f8a61ad5b0a7693ae08154060b790

It looks like we could get away by just committing the complex compound type to file as nectdf-c/netcdf4-python does instead of relying on the transient hdf5 type. This could also be considered as bug.

That means we could significantly simplify this PR. We would not have to use the `auto_complex` keyword, since we would write perfectly sane NetCDF4 files. All these files are readable with former netcdf-c/netcdf4-python versions as well as h5netcdf versions. This would be fully backwards compatible.

h5netcdf/core.py
kmuehlbauer: Not needed with latest approach.

h5netcdf/core.py
kmuehlbauer: I've checked again, see https://github.com/h5netcdf/h5netcdf/pull/227#issuecomment-1902243420

h5netcdf/core.py
```diff
@@ -724,6 +748,23 @@ def _create_child_variable(
...
+            if not isinstance(dtype, CompoundType):
+                # todo check compound types for
+                #  existing complex types which may be used
+                itemsize = np.dtype(dtype).itemsize
+                try:
+                    width = {8: "FLOAT", 16: "DOUBLE"}[itemsize]
+                except KeyError:
+                    raise TypeError(
+                        "Currently only 'complex64' and 'complex128' dtypes are allowed."
+                    )
```
shoyer, line 762: This results in a slightly better traceback ("direct cause of" rather than "during the handling of")
```suggestion
                except KeyError as e:
                    raise TypeError(
                        "Currently only 'complex64' and 'complex128' dtypes are allowed."
                    ) from e
```

h5netcdf/core.py
```diff
@@ -842,7 +882,7 @@ def _create_child_variable(
         # when a variable is first written to, after variable creation.
         # Here we just attach it to every variable on creation.
         # Todo: get this consistent with netcdf-c/netcdf4-python
-        variable._ensure_dim_id()
+        # variable._ensure_dim_id()
```
shoyer, line 885: This also looks unrelated.

h5netcdf/core.py
```diff
@@ -246,7 +259,7 @@ def _attach_coords(self):
             [self._parent._all_dimensions[d]._dimid for d in dims],
             "int32",
         )
-        if len(coord_ids) > 1:
+        if len(coord_ids) >= 1:
```
shoyer, line 262: This looks unrelated

kmuehlbauer: Thanks!

h5netcdf/core.py
kmuehlbauer: Thanks!

kmuehlbauer:
@shoyer Thanks for the immediate review. I've fixed the mentioned issues and applied your change for exception handling.

Sure, xarray's h5netcdf tests (https://github.com/h5netcdf/h5netcdf/actions/runs/7597039010/job/20691426586?pr=227) are breaking, because of the complex compatibility tests over there. This will immediately lead to breaking nightly tests over at xarray, if this PR is merged. I'm ready to take action over at xarray.

kmuehlbauer:
Unfortunately there are still some inconsistencies in the correct retrieval of the committed type. I'll push more changes later.

h5netcdf/core.py
```diff
@@ -724,6 +748,26 @@ def _create_child_variable(
...
+                        "Currently only 'complex64' and 'complex128' dtypes are allowed."
+                    ) from e
+                dname = f"_PFNC_{width}_COMPLEX_TYPE"
+                dtype = self._all_cmptypes.get(
+                    dname, self.create_cmptype(np.dtype(dtype), dname)
+                )
+            # use committed type
+            # that will actually reference the named type in the dataset
+            dtype = dtype._h5ds
+
```
kmuehlbauer, line 770: This actually references the committed type within the dataset. I've used `h5ls` to check this. This tool was recommended at hdf5forum. For some reason I can't retrieve the `name` of the committed type from the dataset datatype. I've struggled to retrieve this from the plain c code over at hdf5 but that did not work out. 

There is one solution to check the dataset type with any committed type for equality, but this seems quite clunky. I'll implement that solution to move forward here.

kmuehlbauer: Just before giving up, I've found another way of retrieving the name of the linked type directly from the dataset (here `datac2`):

```python
ds._root._h5py.h5i.get_name(datac2._h5ds.id.get_type()).decode()
```

Getting back to this the next days. @shoyer Sorry, but this is more involved than I anticipated.

h5netcdf/core.py
```diff
@@ -258,6 +271,28 @@ def _ensure_dim_id(self):
             if "_Netcdf4Dimid" in dim.attrs:
                 self._h5ds.attrs["_Netcdf4Dimid"] = dim.attrs["_Netcdf4Dimid"]
 
+    def _add_fillvalue(self, fillvalue, dtype):
```
kmuehlbauer, line 274: Refactored out from create_child_variable

h5netcdf/core.py
```diff
@@ -316,6 +351,27 @@ def __len__(self):
     @property
     def datatype(self):
         """Return numpy dtype or user defined type."""
+        if self._root._h5py.__name__ == "h5py":
```
kmuehlbauer, line 354: This is the current procedure to extract the committed type as h5netcdf type. For h5pyd and transient types special handling is needed.

h5netcdf/core.py
```diff
@@ -524,10 +582,93 @@ def _unlabeled_dimension_mix(h5py_dataset):
     return status
 
 
+def _check_h5type(self, dtype, np_dtype):
```
kmuehlbauer, line 585: Refactored out from create_child_variable.

h5netcdf/core.py
```diff
@@ -524,10 +582,93 @@ def _unlabeled_dimension_mix(h5py_dataset):
...
+        # create and commit type in current group
+        if dname not in self._all_cmptypes:
+            self.create_cmptype(np_dtype, dname)
+        # get committed type from file
+        return self._all_cmptypes[dname]._h5ds
+    else:
+        return dtype
+
+
+def _check_fillvalue(self, fillvalue, dtype, np_dtype):
```
kmuehlbauer, line 638: Refactored out from create_child_variable

h5netcdf/core.py
```diff
@@ -790,32 +956,21 @@ def _create_child_variable(
...
+        # use numpy dtype for h5pyd
+        if self._root._h5py.__name__ == "h5pyd":
+            h5type = np_dtype
 
         # create hdf5 variable
         self._h5group.create_dataset(
             h5name,
             shape,
-            dtype=dtype,
+            dtype=h5type,
```
kmuehlbauer, line 970: Using the h5type (for user types this would be the committed type) this will link/share the type in the dataset.

h5netcdf/core.py
```diff
@@ -258,6 +271,28 @@ def _ensure_dim_id(self):
...
+                and string_info.length is not None
+                and string_info.length > 1
+            ) or enum_info:
+                value = fillvalue
+            else:
+                value = self.dtype.type(fillvalue)
+
+        # need to use create-function in order
+        # to provide correct committed/named type
+        self.attrs._h5attrs.create("_FillValue", value, dtype=dtype)
```
kmuehlbauer, line 294: Specifying the dtype here (for user types this would be the committed type) will link/share the type in the attribute.

kmuehlbauer (commented):
I've added some comments for better understanding.

h5netcdf/core.py
kmuehlbauer: Note to self: Might need to special case `_FillValue` in `attrs.py` `__setitem__` to write it with dtype (committed type) from variable in any case (user might overwrite `_FillValue` attribute)

kmuehlbauer:
This seems to work now as intended. I'll split out the changes regarding EnumType and shared user type into a separate PR's. This will finally wrap up VLType and CompoundType.

h5netcdf/core.py
```diff
@@ -578,14 +592,31 @@ def _check_dtype(self, dtype):
         # check if committed type can be accessed in current group hierarchy
         dname = dtype.name.split("/")[-1]
         if (
-            (user_type := self._all_enumtypes.get(dname)) is None
-        ) or user_type._h5ds.name != h5type.name:
+            (user_type := self._all_usertypes.get(dname)) is None
+        ) or self._root._h5file[user_type].name != h5type.name:
```
kmuehlbauer, line 596: This looks very ugly. I'll think about another approach to retrieve this.

h5netcdf/core.py
```diff
@@ -691,15 +724,25 @@ def __init__(self, parent, name):
 
         self._dimensions = Dimensions(self)
         self._enumtypes = _LazyObjectLookup(self, self._enumtype_cls)
+        self._vltypes = _LazyObjectLookup(self, self._vltype_cls)
+        self._cmptypes = _LazyObjectLookup(self, self._cmptype_cls)
+        self._usertypes = dict()
```
kmuehlbauer, line 729: Unfortunately we can't utilize _LazyObjectLookup in a neat manner for a joined usertype-dict. Using a simple dict for now, just adding name/h5name key/value-pairs.

h5netcdf/core.py
```diff
@@ -713,11 +756,14 @@ def __init__(self, parent, name):
...
-            ) and self._root._h5py.check_enum_dtype(v.dtype):
-                self._enumtypes.add(k)
+            elif isinstance(v, self._root._h5py.Datatype):
+                if self._root._h5py.check_enum_dtype(v.dtype):
+                    self._enumtypes.add(k)
+                elif self._root._h5py.check_vlen_dtype(v.dtype):
+                    self._vltypes.add(k)
+                elif v.dtype.names is not None or "complex" in v.dtype.name:
+                    self._cmptypes.add(k)
+                self._usertypes[k] = v.name
```
kmuehlbauer, line 766: If someone has a nice idea how to remove that chain of checks...

h5netcdf/core.py
```diff
@@ -1175,9 +1233,44 @@ def create_enumtype(self, datatype, datatype_name, enum_dict):
         self._h5group[datatype_name] = et
         # create enumtype class instance
         enumtype = self._enumtype_cls(self, datatype_name)
+        # enumtype = self._usertype_cls(self, datatype_name)
         self._enumtypes[datatype_name] = enumtype
+        self._usertypes[datatype_name] = enumtype._h5ds.name
```
kmuehlbauer, line 1238: Adding the type to lazy-dict and the h5name to usertype-dict. Same thing below.

h5netcdf/core.py
```diff
@@ -578,14 +592,31 @@ def _check_dtype(self, dtype):
...
+            ) from e
+        dname = f"_PFNC_{width}_COMPLEX_TYPE"
+        # todo check compound type for existing complex types
+        #  which may be used her
+        # if dname is not available in current group-path
+        # create and commit type in current group
+        if dname not in self._all_cmptypes:
+            dtype = self.create_cmptype(dtype, dname).dtype
+        # get committed type from file
+        # return self._all_cmptypes[dname]._h5ds
```
kmuehlbauer, line 619: This is the complex compound creation part. This is really backwards compatible, as h5netcdf always reads these complex compound types as complex numbers. netcdf-c/netcdf4-python can read those compounds if, like we do now, the complex compound type is committed to the file.

kmuehlbauer (commented):
@shoyer @ZedThree

After #229 and #230 (some refactoring to keep good review experience), I've updated this PR with the undisputed changes to get VL and Compound types working for h5netcdf. Plays well together with nc-complex AFAICT.

I'll do some more clean-up the next days, but you might already have a look here.

h5netcdf/core.py
kmuehlbauer: Fixed!

h5netcdf/core.py
kmuehlbauer: Fixed!

h5netcdf/core.py
kmuehlbauer: Fixed!

h5netcdf/core.py
kmuehlbauer: Fixed! Not needed anymore!

kmuehlbauer:
I think this is ready for final review. I really like to get this in sometime next week, and get a release out.

kmuehlbauer:
So, this is unfortunately not finished yet. There are issues with compound type containing strings. This is done by array of chars under the hood, which needs some special casing to be transparent in h5netcdf. See: https://github.com/Unidata/netcdf4-python/issues/773

I'll set this PR to draft until this is figured out.

kmuehlbauer:
@ZedThree @shoyer Finally I got around to finish this up. Please have a look if you are around.

I'm having another self-review early next week and will merge after that.

ZedThree:
LGTM! Thanks @kmuehlbauer !

PR #245, opened 2024-11-10
fix hsds/h5pyd tests

kmuehlbauer:
- [ ] Closes Issue #xxxx
- [ ] Tests added


kmuehlbauer:
@ajelenak I'm having a hard time to find the issue here. 

The tests have silently be broken for some months now, as pip installing `hsds`/`h5pyd` didn't work due to some dependency issue.

In this PR I'm again using the latest versions from respective main. But our old `convtest.py` setup does not work anymore. I hope you can spare some time to look into this, as I'm somehow stuck now. Thanks!

ajelenak:
Thanks for reporting, I'll take a look.

kmuehlbauer:
@ajelenak @jreadey Thanks for the great work over in hsds/h5pyd :100: :1st_place_medal:.

I've brought most of the tests back to work by aligning with the CI setup in hsds/h5pyd. This also simplified the whole test setup for the remote tests, which is really nice!

It would be great if you could have a look into the failing tests, here. There is one test which covers complex numbers, one which checks `track_order` and two tests where there is an issue with reopening the file. Thanks a bunch!

https://github.com/h5netcdf/h5netcdf/actions/runs/12195688403/job/34021801227?pr=245

jreadey:
I was looking at the track_order failure today.  The track_order argument was added to create_group in h5pyd in the last release, but we don't have support for setting track_order by property list.  I'll look into adding this to h5pyd.

I'll take a look at the file re-open and complex failures.

kmuehlbauer:
@jreadey I really appreciate your efforts, here and in #247.

I'm not sure which way of running is more appropriate, the Python app-version via conftest.py  (as of now and #247) or by running hsds completely externally (like here).

I like the current approach (since it's more user friendly wrt test-running) so it would be best to continue the efforts in #247. Anyway, please ping me if I can assist.

jreadey:
I've pushed out a new h5pyd release (v0.20.0) that should fix the track_order issue.  Please re-run the CI tests and let's see how we are doing.

kmuehlbauer:
@jreadey Thanks for working on this! I've re-run the CI here and in #247. It turns out that your #247 is better suited for merge as you've already fixed the two roundtrip tests there.

The remaining issue is with the complex data which is returned as:

```python
<HDF5 dataset "data": shape (5,), type "|V16">
```

with dtype `[('r', '<f8'), ('i', '<f8')]`

`h5py` returns this as 

```python
<HDF5 dataset "data": shape (5,), type "<c16">
```

with dtype `complex128`.

Would it be possible to do the transform from structured array to complex on the h5pyd-side?

jreadey:
That last issue will take some time.  I need to update the type handling in both HSDS and h5pyd to support complex types.  Can we disable this last failure for now?  
I've added an issue in h5pyd for complex support: https://github.com/HDFGroup/h5pyd/issues/246.

kmuehlbauer:
Sure, let's just skip the test for h5pyd/hands. Thanks @jreadey!

kmuehlbauer:
@jreadey I'm closing this PR in favour of your #247.

Issue #240, opened 2024-10-02
zstd support?


hmaarrfk:
Does h5netcdf support zstd compression?

#### Is your feature request related to a problem? Please describe.
```
import xarray as xr
ds = xr.DataArray(np.zeros(3), name='data).to_dataset()
ds['data'].encoding['compression'] = 'zstd'
# netcdf4 works
ds.to_netcdf('test.nc', engine='netcdf4')

# h5netcdf doesn't
ds.to_netcdf('test.nc', engine='h5netcdf')
```
```
ValueError: Compression filter "zstd" is unavailable
```

#### Describe the solution you'd like
Just to know if there is a path forward at all...

#### Describe alternatives you've considered
netcdf4. But we've found that it has some strange bugs in our workflows....

#### Additional context
I can help compile things on conda-forge.

kmuehlbauer:
@hmaarrfk `h5netcdf` should already be able to use [`hdf5plugin`](http://www.silx.org/doc/hdf5plugin/latest/) out of the box. How about:

```python
import numpy as np
import xarray as xr
import hdf5plugin
ds = xr.DataArray(np.zeros(3), name='data').to_dataset()
ds['data'].encoding['compression'] = hdf5plugin.Zstd()
# h5netcdf works
ds.to_netcdf('test.nc', engine='h5netcdf')
# netcdf4 doesn't
ds.to_netcdf('test.nc', engine='netcdf4')
```

Seems to read just fine for both, so I'm assuming there is some incantation to make `hdf5plugin` usable for `netcdf4-python` on write the same way.

```python
with xr.open_dataset("test.nc", engine="h5netcdf") as h5ds:
    with xr.open_dataset("test.nc", engine="netcdf4") as ncds:
        xr.testing.assert_equal(h5ds, ncds)
```

hmaarrfk:
Hmm that does seem to look like it works. But it is definitely "clunky" in the sense that I must know what my backend is going to be do to it right.

I guess some PRs could be made to xarray to help tie in the string `"zstd"` to `hdf5plugin.Zstd()` when the h5netcdf plugin is used.

Thanks1

kmuehlbauer:
@hmaarrfk I agree, it would be great to have a unified API. We could also implement the netcdf4-python behavior into h5netcdf until there is a more generic approach.

hmaarrfk:
I think it might be a combination of both.

I would have to study how xarray communicates with netcdf4 more.

[...]

PR #252, opened 2025-01-24
Allow zlib to be an alias for gzip

hmaarrfk:
libnetcdf4 allows zlib to be passed in as a compression string.

In xarray one would typically allow 'zlib' to be specified in

encoding['compression'] = 'zlib'

but the problem is that h5netcdf won't support this syntax.

This builds that compatibility layer

not quite related but: https://github.com/h5netcdf/h5netcdf/issues/240

- [x] Tests added

### DIFF
--- a/h5netcdf/core.py
+++ b/h5netcdf/core.py
@@ -1185,7 +1185,7 @@ class Group(Mapping):
         compression : str, optional
             Compression filter to apply, defaults to ``gzip``. ``zlib`` is an alias for ``gzip``.
         compression_opts : int
-            Parameter for compression filter. For ``compression="gzip"`` Integer from 1 to 9 specifying
+            Parameter for compression filter. For ``compression="gzip"``/``compression="zlib"`` Integer from 1 to 9 specifying
             the compression level. Defaults to 4.
         fletcher32 : bool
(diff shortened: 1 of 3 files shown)


h5netcdf/core.py
```diff
@@ -1183,7 +1183,7 @@ def create_variable(
             ``h5netcdf``. Discussion on ``h5netcdf`` chunking can be found in (:issue:`52`)
             and (:pull:`127`).
         compression : str, optional
-            Compression filter to apply, defaults to ``gzip``
+            Compression filter to apply, defaults to ``gzip``. ``zlib`` is an alias for ``gzip``.
         compression_opts : int
             Parameter for compression filter. For ``compression="gzip"`` Integer from 1 to 9 specifying
```
kmuehlbauer, line 1188:
```suggestion
            Parameter for compression filter. For ``compression="gzip"``/``compression="zlib"`` Integer from 1 to 9 specifying
```

kmuehlbauer (approved):
Thanks @hmaarrfk! One suggestion for the docstring to be more explicit. Otherwise this is good to merge.

hmaarrfk:
> One suggestion

feel free to merge similar suggestions. I appreciate your speed

kmuehlbauer:
Thanks @hmaarrfk, much appreciated!

PR #254, opened 2025-01-26
REL: 1.5.0

kmuehlbauer:
cc @hmaarrfk


hmaarrfk:
Wow thanks! This avoids me creating longing rules against this syntax.

kmuehlbauer:
@hmaarrfk Release numbers are cheap :grinning: and your collaboration here is always much appreciated.

PR #258, opened 2025-03-07
REL: 1.6.0

kmuehlbauer:


kmuehlbauer:
@rho-novatron FYI, a new release is on the way.

PR #259, opened 2025-03-07
Return None for chunks on scalar variables

rho-novatron:
- [ ] Closes Issue #xxxx
- [ ] Tests added

### DIFF
--- a/h5netcdf/core.py
+++ b/h5netcdf/core.py
@@ -608,6 +608,9 @@ class BaseVariable(BaseObject):
 class Variable(BaseVariable):
     @property
     def chunks(self):
+        if self.shape == ():
+            assert self._h5ds.chunks in (None, (), (1,))
+            return None
         return self._h5ds.chunks
 
     @property


h5netcdf/core.py
```diff
@@ -608,6 +608,9 @@ def __repr__(self):
 class Variable(BaseVariable):
     @property
     def chunks(self):
+        if self.shape == ():
```
kmuehlbauer, line 611: @rho-novatron Thanks! Please leave a comment here, for better visibility, why we are doing this.

kmuehlbauer (commented):
@rho-novatron Thanks for the fast PR! I think this is the right place for the fix. Would be nice to have a test for this, too.

h5netcdf/core.py
rho-novatron: Done!

rho-novatron: Done

kmuehlbauer:
Would you consider this a (bug)-fix or enhancement? I'm just asking if we need to increase MINOR or PATCH for the next release ;-) Anyway, can I ask you to add a CHANGELOG.rst entry again?

rho-novatron:
Since it's mostly about being consistent in returning what was provided when constructing the dataset (no chunks), and not crashing xarray, I would be completely fine with this being a bug fix. Changelog added! I'm really liking these turnaround times. :-)

kmuehlbauer:
Thanks again for speeding things, up, @rho-novatron!

kmuehlbauer:
> I'm really liking these turnaround times. :-)

It's not always the case :grimacing:, but today seemed a good day for trying to fight against **_Do not deploy on Friday_**! Have a nice weekend!

Issue #262, opened 2025-06-10
File.__init__ with the h5pyd engine constructs the underlying h5pyd.File twice


rho-novatron:
#### What happened:
When constructing a h5netcdf File with the h5pyd engine, the construction is much slower than expected.

This seems to be due to h5netcdf constructing the h5pyd.File once just to check whether the hsds domain exists. Then that constructed h5pyd.File is thrown away, and a new h5pyd.File is constructed. This takes about 1000 ms extra in regular cases for our workload.

#### What you expected to happen:
The simplest fix may be to not check whether the hsds domain already exists when opening with `mode="r"`. That would still leave the slow path in place for when opening with other modes. Perhaps the h5pyd File class could include the information on whether the domain already existed as a property?

#### MCVE Code Sample

See lines 1522-1531 in https://github.com/h5netcdf/h5netcdf/blob/41b53345d3407651a6379844c09163a19d3c925d/h5netcdf/core.py#L1522C21-L1531C22

#### Expected Output
Only one call should be made.

kmuehlbauer:
@rho-novatron Thanks for the report. It looks like the current situation is based on organic growth. 

The idea behind `_preexisting_file` was to not overwrite this attribute in any case. It should only be written if the file was created by `h5netcdf`. This can happen if:

- `mode="w"` or 
- `mode="a"` and the file is not yet existing

kmuehlbauer:
@rho-novatron Thanks again for filing this issue. Using h5pyd the file isn't opened twice any more. Instead, files are opened directly with the requested mode, but:

```python
("Append mode for h5pyd now probes with 'r+' first and "
 "only falls back to 'w' if the file is missing.\n"
 "To silence this warning use 'r+' (open-existing) or 'w' "
 "(create-new) directly."
)
```

Please comment, if that suits your requirements. Thanks!

PR #263, opened 2025-06-12
Avoid opening file with h5pyd just to check if it exists.

JonasGronberg:
- [ ] Closes Issue #262

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,11 @@
 Change Log
 ----------
 
+Development Version (unreleased):
+
+- Avoid opening h5pyd file to check if there is a preexisting file ({issue}`262`).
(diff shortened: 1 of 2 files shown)


kmuehlbauer:
@JonasGronberg Thanks for taking this on. As already mentioned in #262, the solution is not as straightforward as it seems, as we would need to check if the file exists in case of `mode="a"`.

kmuehlbauer:
@JonasGronberg I've just merged #266 with some additions to your initial setup. Working through this here wasn't possible due to lack of permissions on your branch.

Citing from the new UserWarning:

```python
("Append mode for h5pyd now probes with 'r+' first and "
 "only falls back to 'w' if the file is missing.\n"
 "To silence this warning use 'r+' (open-existing) or 'w' "
 "(create-new) directly."
)
```

Thanks for taking the time to contribute, it's much appreciated!

Issue #248, opened 2024-12-22
test_no_circular_references fails with Python 3.14


penguinpee:
#### What happened:

Fedora is running some (very) early tests with upcoming Python 3.14. All Python packages in Fedora are continuously rebuilt against pre-releases of Python 3.14 to signal issues early. Below test failure popped up.

```python
=================================== FAILURES ===================================
___________________ test_no_circular_references[testfile.nc] ___________________

tmp_local_or_remote_netcdf = '/tmp/pytest-of-mockbuild/pytest-0/test_no_circular_references_te0/testfile.nc'

    def test_no_circular_references(tmp_local_or_remote_netcdf):
        # https://github.com/h5py/h5py/issues/2019
        with h5netcdf.File(tmp_local_or_remote_netcdf, "w") as ds:
            ds.dimensions["x"] = 2
            ds.dimensions["y"] = 2
    
        gc.collect()
        with h5netcdf.File(tmp_local_or_remote_netcdf, "r") as ds:
            refs = gc.get_referrers(ds)
            for ref in refs:
                print(ref)
>           assert len(refs) == 1
E           assert 0 == 1
E            +  where 0 = len([])

h5netcdf/tests/test_h5netcdf.py:1551: AssertionError
```

#### Anything else we need to know?:

I browsed through the referenced `h5py` issue, but it didn't help me understand what might be the issue here. Neither did [Python's changelog](https://docs.python.org/3.14/whatsnew/3.14.html). `h5py` was rebuilt successfully in the same build environment.

#### Version

<details><summary>Output of <tt>print(h5py.version.info, f"\nh5netcdf {h5netcdf.__version__}")</tt></summary>

Summary of the h5py configuration
---------------------------------

h5py    3.12.1
HDF5    1.14.5
Python  3.14.0a3 (main, Dec 17 2024, 00:00:00) [GCC 14.2.1 20241104 (Red Hat 14.2.1-6)]
sys.platform    linux
sys.maxsize     9223372036854775807
numpy   2.2.1
cython (built with) 3.0.11
numpy (built against) 2.2.0
HDF5 (built against) 1.14.5
 
h5netcdf 1.4.1

</details>

kmuehlbauer:
Thanks @penguinpee for bringing this to attention. 

I've introduced tests for 3.13 over in #250 but no issues there. So it's really something with Python 3.14. If you have any idea how to setup a CI run with Python 3.14 prereleases please let me know.

[...]

PR #264, opened 2025-06-21
Fix and add circular referrer tests for Python 3.14

kmuehlbauer:
- [x] Closes Issue #248 
- [x] Tests added

### DIFF
--- a/h5netcdf/tests/test_h5netcdf.py
+++ b/h5netcdf/tests/test_h5netcdf.py
@@ -1551,9 +1551,6 @@ def test_dimensions(tmp_local_netcdf, read_write_matrix):
     )
 
 
-@pytest.mark.skipif(
-    python_version >= version.parse("3.14"), reason="Only works for Python < 3.14"
-)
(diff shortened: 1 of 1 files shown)


kmuehlbauer:
@TomAugspurger In #117 back in 2021 you've fixed an issue with circular imports. The test added back then is failing for Python 3.14, because there have been changes to the garbage collection. I've tried to adapt this here. It would be great If you could review here to see if the new test does what it should. Thanks!

TomAugspurger:
Looks like this could be https://github.com/python/cpython/issues/125603.

I'm not 100% sure, but maybe the existing test could be changed to `assert len(refs) <= 1` instead of `== 1`, or perhaps `==  0` for python 3.14 and later. I think that'd be consistent with the test, that no one else has a reference to the dataset object (assuming the new `gc.get_referrers` behaves like I'm expecting).

kmuehlbauer:
Great! Many thanks @TomAugspurger also for the linked issue! So I'm going to enable the test with checking == 0 for py314, then. I'll keep the additional test, maybe it will useful in the future.

PR #272, opened 2025-07-12
cleanup: pyupgrade --py39-plus

schwehr:
`pre-commit run --all-files` also removed a blank like from bug_report.md.


kmuehlbauer (approved):
Thanks @schwehr!

Issue #287, opened 2025-10-06
Unexpected resizing with slices larger than data size


Thomas-Z:
#### What happened:

It appears that h5netcdf incorrectly triggers a resize operation when reading data using slices larger than the underlying array.

#### What you expected to happen:

The data should remain unaltered, no resizing should occur, regardless of the slice size.

#### MCVE Code Sample

```python
import h5netcdf
import netCDF4
import numpy as np

with h5netcdf.File("test.nc", "w") as f:
    f.dimensions = {"x": 5}
    v = f.create_variable("longitude", ("x",), data=np.arange(5))

with h5netcdf.File("test.nc", mode="r") as ds:
    lon = ds.variables["longitude"]
    print(lon[slice(0, 10)][:])

# > [0 1 2 3 4 0 0 0 0 0]

with netCDF4.Dataset("test.nc", mode="r") as ds:
    lon = ds.variables["longitude"]
    print(lon[slice(0, 10)][:])

# > [0 1 2 3 4]
```

#### Expected Output

# > [0 1 2 3 4]

#### Version

<details><summary>Output of <tt>print(h5py.version.info, f"\nh5netcdf {h5netcdf.__version__}")</tt></summary>

Summary of the h5py configuration
---------------------------------

h5py    3.14.0
HDF5    1.14.6
Python  3.13.7 | packaged by conda-forge | (main, Sep  3 2025, 14:30:35) [GCC 14.3.0]
sys.platform    linux
sys.maxsize     9223372036854775807
numpy   2.3.3
cython (built with) 3.1.3
numpy (built against) 2.3.2
HDF5 (built against) 1.14.6
 
h5netcdf 1.6.4

</details>

PR #288, opened 2025-10-06
do not return padded arrays if slice is exceeding variable shape

kmuehlbauer:
- [x] Closes Issue #287
- [x] Tests added

### DIFF
--- a/h5netcdf/tests/test_h5netcdf.py
+++ b/h5netcdf/tests/test_h5netcdf.py
@@ -1306,6 +1306,7 @@ def test_reading_unlimited_dimensions_created_with_c_api(tmp_local_netcdf):
         assert f["dummy5"].shape == (2, 3)
 
         # regression test for https://github.com/pydata/xarray/issues/10818
+        # h5netcdf issue https://github.com/h5netcdf/h5netcdf/issues/287
         # slicing a variable with slice exceeding it's shape
         # should return only up to shape-size
         assert f["dummy1"][:10, :2].shape == (2, 2)


h5netcdf/core.py
```diff
@@ -569,8 +570,11 @@ def __getitem__(self, key):
             if string_info and string_info.length is None:
                 return self._h5ds.asstr()[key]
 
-        # get padding
+        # get
+        print("key:", key)
+        # print("size", self.size)
         padding = self._get_padding(key)
+        print("padding:", padding)
```
kmuehlbauer, line 577:
```suggestion
        # get padding
        padding = self._get_padding(key)
```

h5netcdf/tests/test_h5netcdf.py
```diff
@@ -1305,6 +1305,11 @@ def test_reading_unlimited_dimensions_created_with_c_api(tmp_local_netcdf):
         f.groups["test"]["dummy4"].shape == (0, 0)
         assert f["dummy5"].shape == (2, 3)
 
+        # regression test for https://github.com/pydata/xarray/issues/10818
```
kmuehlbauer, line 1308:
```suggestion
        # regression test for https://github.com/pydata/xarray/issues/10818
        # h5netcdf issue https://github.com/h5netcdf/h5netcdf/issues/287
```

kmuehlbauer:
I wonder why this hasn't come up earlier. As this has been detected under the influence of `xarray` changing the default engine to `h5netcdf`, I'm expecting more issues to pile up when users choose `h5netcdf` as engine in `xarray`. But let's see.

I'll get this in later today.

PR #279, opened 2025-08-15
fixup testsuite and issues to work with pyfive

kmuehlbauer:
- [ ] Closes Issue #xxxx
- [ ] Tests added
- [ ] Changes are documented in `CHANGELOG.rst`

This is on top of #273. @davidhassel please have a look here. More tests need to be adapted to use the new backend scheme. But I thought this would already be helpful at this early stage.

This would need the changes in https://github.com/NCAS-CMS/pyfive/pull/83/files to work properly.

PR #304, opened 2025-11-23
add new backend code, prepare test suite for future pyfive addition

kmuehlbauer:
This extracts the code which was added in the course of #279 to make h5netcdf capable of parametrizing the used backend. The testsuite was also prepared for this change.  

- [x] Tests were added/refactored


kmuehlbauer:
Nevermind, I thought we can split #279 into two. But it's not worth the effort.

Issue #310, opened 2026-01-19
Missing Dependencies - numpy and h5py


bkupiak:
#### What happened:

Upon installing h5netcdf v1.8.0 and trying to import the package, I received `ModuleNotFoundError: No module named 'numpy'`

Installed numpy and then tried importing h5netcdf again and received `ModuleNotFoundError: No module named 'h5py`

#### What you expected to happen:

Expected h5netcdf to automatically install all its dependencies and successfully be imported without needing to explicitly install additional required packages.

#### MCVE Code Sample

To reproduce `ModuleNotFoundError: No module named 'numpy'`:
```docker run -it python:3.10-bullseye bash -c "pip install h5netcdf && python -c 'import h5netcdf'"```

To reproduce `ModuleNotFoundError: No module named 'h5py`:
```docker run -it python:3.10-bullseye bash -c "pip install h5netcdf numpy && python -c 'import h5netcdf'"```

#### Expected Output

No python output (no ModuleNotFoundError)

#### Version

h5netcdf v1.8.0 installed

<details><summary>Output of <tt>print(h5py.version.info, f"\nh5netcdf {h5netcdf.__version__}")</tt></summary>

N/A - import unsuccessful

</details>

kmuehlbauer:
@bkupiak Can you install with

`pip install h5netcdf[h5py]`

If that works I need to find out if a fallback can be defined, that it always installs h5py if no option is provided.

bkupiak:
Yes, this works! `docker run -it python:3.10-bullseye bash -c "pip install h5netcdf[h5py] && python -c 'import h5netcdf'"`

kmuehlbauer:
OK, the adoption of `pyfive` moved `h5netcdf` in a new situation:

1. Ability to install `pyfive` and `h5py` separately:

    ```
    pip install h5netcdf[h5py]
    pip install h5netcdf[pyfive]
    ```

  This works nicely.

2. Ability to do and install a default dependency (eg. h5py):

    ```
    pip install h5netcdf
    ```

  This doesn't work atm. It would work with [PEP 771](https://peps.python.org/pep-0771/), IIUC.

Many libraries solved this by raising an import error on missing dependencies, and requesting an install. They also document this behaviour in their install section.

@bkupiak Would this solve your issue with the above outlined additions?

bkupiak:
[...]

PR #311, opened 2026-01-21
fix backend dependencies

kmuehlbauer:
fix backend dependencies, update install docs, add more descriptive error mes…sage on missing backend

- [x] Closes Issue #310

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,10 @@
 Change Log
 ----------
 
+Version 1.8.1 (January Nst, 2026):
+
+- Add numpy to default dependencies, update install section to highlight backend installation, update error message on missing backend (:issue:`310`, :pull:`311`) by `Kai Mühlbauer <https://github.com/kmuehlbauer>`_
+
 Version 1.8.0 (January 16st, 2026):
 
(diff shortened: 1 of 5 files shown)


bkupiak:
This looks like a reasonable workaround to solve #310. As mentioned in the issue, an ideal solution would be to include `h5py` as a required dependency (just like `packaging` and `numpy`), but I understand that is not an option currently. @kmuehlbauer,  for the sake of documentation and my own curiosity, could you explain why that is not possible?

kmuehlbauer:
First, this PR need still some fixes. The test suite still needs some overhaul.

@bkupiak I'd happily implement a solution if there is any. With the adoption of `pyfive` as drop-in replacement for `h5py`, users should be able and would want to install `h5netcdf` with `pyfive` INSTEAD of `h5py`. This will not work, if we have `h5py` as required dependency.

The mentioned [PEP771](https://peps.python.org/pep-0771/) aims to provide exactly this functionality, having a default dependency (`h5py` in our case) to `pip install h5netcdf` with this default and also be able to install `pyfive` `pip install h5netcdf[pyfive]` without installing `h5py`. Does that make sense?

bkupiak:
Ah, yes, I see, thank you!

